PR #3, opened 2021-11-03
Use top-level JSON properties when deserializing cluster detail

brandur:
Traditionally, cluster details were wrapped in a `cluster:{}` subobject,
but we've since moved to preferring top-level properties instead. We
still return the subobject, but are trying to deprecate it.

Here we change CB to read properties off of the top-level JSON object.


brandur:
@will Can you check this? Thanks!

PR #5, opened 2022-01-18
Add `--database` flag to `scope` and `psql`

abrightwell:
These changes add the ability to specify a database name to connect when
using the `scope` and `psql` command.

The flag is setup to be optional. If not provided then it will default
to `postgres`.

```
> ./bin/cb psql --help
Usage: cb psql <cluster id> [--database] [-- [args for psql such as -c or -f]]
    --database NAME                  Database name (default: postgres)

> ./bin/cb scope --help
Usage: cb scope <--cluster> [--(check),...]
...
    --database NAME                  Database name (default: postgres)
...
```

Example with `psql`:

```
> ./bin/cb psql <test-cluster-id>
connecting to personal/test-cluster
psql (14.1 (Ubuntu 14.1-1.pgdg18.04+1), server 13.4)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

personal/test-cluster/postgres=# \q

> ./bin/cb psql <test-cluster-id> --database foo
connecting to personal/test-cluster
psql (14.1 (Ubuntu 14.1-1.pgdg18.04+1), server 13.4)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

personal/test-cluster/foo=# \q
```

### DIFF
--- a/spec/cb/completion_spec.cr
+++ b/spec/cb/completion_spec.cr
@@ -349,6 +349,7 @@ describe CB::Completion do
     result.should have_option "--suite"
     result.should have_option "--mandelbrot"
     result.should have_option "--tables"
+    result.should have_option "--database"
 
     result = parse("cb scope --cluster abc --suite ")
     result.should have_option "all"
@@ -362,5 +363,8 @@ describe CB::Completion do
     result = parse("cb scope --cluster abc --mandelbrot ")
     result.should have_option "--tables"
     result.should_not have_option "--mandelbrot"
+
+    result = parse("cb scope --cluster abc --database ")
+    result.should_not have_option "--database"
   end
 end
--- a/src/cb/psql.cr
+++ b/src/cb/psql.cr
@@ -2,10 +2,13 @@ require "./action"
 
 class CB::Psql < CB::Action
   property cluster_id : String?
+  property database : String?
 
   def run
     c = client.get_cluster cluster_id
     uri = client.get_cluster_default_role(cluster_id).uri
+
+    database.tap { |db| uri.path = db if db }
 
     output << "connecting to "
     team_name = print_team_slash_cluster c, output
@@ -30,6 +33,10 @@ class CB::Psql < CB::Action
   def cluster_id=(str : String)
     raise_arg_error "cluster id", str unless str =~ EID_PATTERN
     @cluster_id = str
+  end
+
+  def database=(str : String)
+    @database = str
   end
 
   private def ensure_cert(team_id) : String
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -77,8 +77,10 @@ op = OptionParser.new do |parser|
   end
 
   parser.on("psql", "Connect to the database using `psql`") do
-    parser.banner = "Usage: cb psql <cluster id> [-- [args for psql such as -c or -f]]"
+    parser.banner = "Usage: cb psql <cluster id> [--database] [-- [args for psql such as -c or -f]]"
     action = psql = CB::Psql.new(PROG.client)
+
+    parser.on("--database NAME", "Database name (default: postgres)") { |arg| psql.database = arg }
 
     parser.unknown_args do |args|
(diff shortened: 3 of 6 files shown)


spec/cb/completion_spec.cr
```diff
@@ -102,7 +102,7 @@ describe CB::Completion do
     result.should_not have_option "--replica"
 
     result = parse("cb create --fork abc --at ")
-    result.should eq ([] of String)
```
abrightwell, line 105: The changes in this file were just from running `crystal tool format`.

src/cb/psql.cr
```diff
@@ -19,7 +20,7 @@ class CB::Psql < CB::Action
       "PGHOST"        => uri.hostname,
       "PGUSER"        => uri.user,
       "PGPASSWORD"    => uri.password,
-      "PGDATABASE"    => uri.path.lchop('/'),
+      "PGDATABASE"    => database.nil? ? uri.path.lchop('/') : database,
```
abrightwell, line 23: Not sure if this is desirable or doing like was done in `scope` and just reassigning `uri.path` if the flag was set.

spec/cb/completion_spec.cr
will: I moved this to it's own commit since I merged some of the other things first, so we can remove these changes from this PR

src/cb/psql.cr
```diff
@@ -32,6 +37,11 @@ class CB::Psql < CB::Action
     @cluster_id = str
   end
 
+  def database=(str : String)
+    str = str.downcase
```
will, line 41: Do we want this? I don’t know what happens if people have upper case characters in their database name with postgres at all.

src/cli.cr
```diff
@@ -77,9 +77,11 @@ op = OptionParser.new do |parser|
   end
 
   parser.on("psql", "Connect to the database using `psql`") do
-    parser.banner = "Usage: cb psql <cluster id> [-- [args for psql such as -c or -f]]"
+    parser.banner = "Usage: cb psql <cluster id> [<--database>] [-- [args for psql such as -c or -f]]"
```
will, line 80:
```suggestion
    parser.banner = "Usage: cb psql <cluster id> [--database] [-- [args for psql such as -c or -f]]"
```

I've been going with `[]` is for optional and `<>` for required, so I think we just need the `[]`

src/cb/psql.cr
```diff
@@ -2,11 +2,16 @@ require "./action"
...
   property cluster_id : String?
+  property database : String?
 
   def run
     c = client.get_cluster cluster_id
     uri = client.get_cluster_default_role(cluster_id).uri
 
+    if database.presence
+      uri.path = database.to_s
+    end
```
will, line 13: This is the same reason as the case in the other PR with the instance variable still being a union with nil even in the conditional. It is technically possible for some thread to change the value of `@database` (which these two call to `#database` is returning) to change between calls. For this cli program, it'll never happen, but that's why the call to `#presence` really isn't doing much, and why I think probably the complier made you tack on `#to_s`.



One way would be 
```suggestion
    if (db = database)
      uri.path = db
    end
```

But also you could do it I think like this, which is nice because the above creates a local variable that is in scope for the rest of the method, whereas the block argument is only in scope in the block.

```suggestion
    database.tap { |db| uri.path = db if db }
```

I don’t think it matters a whole lot either way. Just explaining all the options and what's going on assuming you're new to Crystal.

will (changes requested):
Could you please also add completion and completion specs, and a changelog entry?

src/cb/psql.cr
abrightwell: So, yeah, it's likely not necessary, but testing `foo` vs `FOO` it does the right thing. So really should matter, will remove.

PR #6, opened 2022-01-19
Refactor team `roles` to use `role` from response.

abrightwell:
These changes refactor the `teams` command to utilize the `role` field
found in the `teams` response as `roles` is no deprecated.

Example:

```
> ./bin/cb teams
<personal_id>      personal
<team-1-id>      adam-test-team                  Member, Manager, Administrator
<team-2-id>      <script>alert(1)</script>       Member, Manager, Administrator

> ./bin/cb teams
<personal_id>      personal
<team-1-id>      adam-test-team                  Admin
<team-2-id>      <script>alert(1)</script>       Admin
```

### DIFF
--- a/src/cb/client.cr
+++ b/src/cb/client.cr
@@ -88,19 +88,9 @@ class CB::Client
     Account.from_json resp.body
   end
 
-  jrecord Team, id : String, name : String, is_personal : Bool, roles : Array(Int32) do
-    enum Role
-      Member
-      Manager
-      Administrator
-    end
-
+  jrecord Team, id : String, name : String, is_personal : Bool, role : String? do
     def name
       is_personal ? "personal" : @name
-    end
-
-    def human_roles
-      roles.map { |i| Role.new i }
     end
   end
 
--- a/src/cb/program.cr
+++ b/src/cb/program.cr
@@ -100,7 +100,7 @@ class CB::Program
       output << "\t"
       output << team.name.ljust(name_max).colorize.t_name
       output << "\t"
-      output << team.human_roles.join ", "
+      output << team.role.to_s.titleize
       output << "\n"
     end
   end
--- a/spec/cb/completion_spec.cr
+++ b/spec/cb/completion_spec.cr
@@ -6,7 +6,7 @@ private class CompletionTestClient < CB::Client
   end
 
   def get_teams
-    [Team.new("def", "my team", false, [1])]
+    [Team.new("def", "my team", false, "manager")]
   end
 
   def get_firewall_rules(id)


src/cb/client.cr
```diff
@@ -88,19 +88,16 @@ class CB::Client
...
       is_personal ? "personal" : @name
     end
 
-    def human_roles
-      roles.map { |i| Role.new i }
+    def role
+      if @role.presence
+        @role = (@role == "admin") ? "Administrator" : @role
+        @role.as(String).titleize
+      end
```
abrightwell, line 100: I was hoping this could be a bit simpler and on the surface I felt like it could be. However, since `role` is `String | Nil` it seemed like it was necessary (or at least the compiler made it seem that way :shrug:). Is there a better way around this?

will: `role` is a union of `String` and `Nil` because you defined it as a `role : String?`. The question mark is a shortcut meaning the same thing. Also the return type of this method is `String?` because if the conditional evaluates to false, the if block returns `nil` otherwise a string.

You might have thought that the check to `@role.presence` would mean that inside that conditional the complier would know that `@role` is not `nil`. However that assumption isn't thread-safe, since each access to an instance variable could have been changed from another thread. So that's why it's still `String?` inside the conditional. You could store `@role` in a local variable and do the conditional on that instead since the local variable wont change.

However, if we know that the api will always return a string, we can drop the `?` there and it'd be only of type `String`.  This would also remove the conditional, and change the return type of the method to just `String` which is nice.

As it stands now though, if the api can actually return a nil/null for that field though, that cast with `as` will fail: https://play.crystal-lang.org/#/r/cnp9 . The local variable trick would fix that.

I'd also be okay with changing the display of the role it to just the word, `admin` if that is what the api returns, then we wouldn’t need the overwrite with "Administrator" conditional.

If we do both [be okay with just `admin`] and [make the type `String` sans `?`], we could move the title casing to when it's displayed in program.cr and actually remove this `#role` method altogether.

will (changes requested):
So now there is only one role instead of a list of roles?

abrightwell:
> However, if we know that the api will always return a string, we can drop the ? there and it'd be only of type String. This would also remove the conditional, and change the return type of the method to just String which is nice.

Yeah, unfortunately, right now `null` is possible if the team is considered `personal`.

> I'd also be okay with changing the display of the role it to just the word, admin if that is what the api returns, then we wouldn’t need the overwrite with "Administrator" conditional.

WFM. Initially my thoughts were simply around preserving current behaviors. But if you're cool with simply `admin` then :+1:.

I'll update accordingly.

abrightwell:
> So now there is only one role instead of a list of roles?

Yes. `roles` is now deprecated in favor of the singular `role`.

will:
> Yeah, unfortunately, right now null is possible if the team is considered personal.

Do you think the api should return admin for that, since you are the admin of your own personal team?

abrightwell:
> Do you think the api should return admin for that, since you are the admin of your own personal team?

That's a good question. @brandur thoughts here?

brandur:
> That's a good question. @brandur thoughts here?

No strong preference from me. Yeah, maybe we should change it if it simplifies the implementation of API consumers like it would seem to here.

src/cb/program.cr
```diff
@@ -100,7 +100,7 @@ class CB::Program
       output << "\t"
       output << team.name.ljust(name_max).colorize.t_name
       output << "\t"
-      output << team.human_roles.join ", "
+      output << team.role.to_s.titleize
```
abrightwell, line 103: How about this for now? Since `nil` will be `""`. That way impact is minimal/negligible regardless of the direction that API takes in the near term.

will:
Thanks, did the rebase on my machine

PR #9, opened 2022-01-20
Fix create cluster with ha issue.

abrightwell:
The request being sent by the client was incorrect to enable ha. This
fixes that issue.


will:
Thanks for finding this issue. If you could also please update the changelog with the same PR, that would make it easier for me not to forget

PR #10, opened 2022-02-18
Add `cb restart` command

abrightwell:
Adds a command to restart clusters.

This command requires either 1) a confirmation of the cluster name to be restarted or 2) the use of the `--confirm` flag when running the command.

Example:

```
> cb restart jlearutdxvaohdxltdbqz3lyoe
About to restart cluster test-cluster-1.
  Type the cluster's name to confirm: test-cluster-1
Cluster jlearutdxvaohdxltdbqz3lyoe restarted

> cb restart jlearutdxvaohdxltdbqz3lyoe --confirm
Cluster jlearutdxvaohdxltdbqz3lyoe restarted

> cb info jlearutdxvaohdxltdbqz3lyoe
personal/test-cluster-1
     state: restarting
...
```

### DIFF
--- a/src/cb/restart.cr
+++ b/src/cb/restart.cr
@@ -1,0 +1,38 @@
+require "./action"
+
+class CB::Restart < CB::Action
+  property cluster_id : String?
+  property confirmed : Bool = false
+
+  def run
+    validate
+
+    c = client.get_cluster cluster_id
+
+    unless confirmed
+      output << "About to " << "restart".colorize.t_warn << " cluster " << c.name.colorize.t_name
+      output << ".\n  Type the cluster's name to confirm: "
+      response = input.gets
+
+      if c.name == response
+        confirmed = true
+      else
+        raise Error.new "Response did not match, did not restart the cluster"
+      end
+    end
+
+    client.restart_cluster cluster_id
+    output.puts "Cluster #{c.id.colorize.t_id} restarted"
+  end
+
+  def validate
+    check_required_args do |missing|
+      missing << "cluster" unless cluster_id
+    end
+  end
+
+  def cluster_id=(str : String)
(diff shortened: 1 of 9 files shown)


pdsphil:
For BR-1010

src/cb/restart.cr
```diff
@@ -0,0 +1,38 @@
...
+class CB::Restart < CB::Action
+  property cluster_id : String?
+  property confirmed : Bool = false
+
+  def run
+    validate
+
+    c = client.get_cluster self.cluster_id
+
+    unless self.confirmed
```
will, line 12:
```suggestion
    c = client.get_cluster cluster_id

    unless confirmed
```

Don't need to explicitly use `self` here. You do when assigning to disambiguate creating a local variable and calling the method `#confirmed=`, but here since there is no local variable with these names, it knows to call the getter method that’s defined when using the `property` macro.

src/cb/restart.cr
```diff
@@ -0,0 +1,38 @@
...
+      response = input.gets
+
+      if c.name == response
+        self.confirmed = true
+      else
+        raise Error.new "Response did not match, did not restart the cluster"
+      end
+    end
+
+    client.restart_cluster @cluster_id
```
will, line 24:
```suggestion
    client.restart_cluster cluster_id
```

Also since you did define the `proeprty` on this, we can use the method here too. It doesn't matter very much at all, but just to keep it the same as above.

will (approved):
Overall very good, thanks for doing the completions too. Just a few comments on implicit `self` and getter/setters vs the direct instance variable.

src/cb/restart.cr
abrightwell: Fixed.

PR #11, opened 2022-02-18
Dont send date in version to sentry

will:
The extra date information made sentry's "mark as fixed in version x"
feature not work right.

### DIFF
--- a/src/cb.cr
+++ b/src/cb.cr
@@ -1,15 +1,15 @@
 module CB
   EID_PATTERN = /\A[a-z0-9]{25}[4aeimquy]\z/
 
-  VERSION       = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
-  BUILD_RELEASE = {{ flag?(:release) }}
-  BUILD_DATE    = {{ `date -u +"%Y%m%d%H%M"`.chomp.stringify }}
-  BUILD_ID      = begin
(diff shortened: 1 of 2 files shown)


will:
anyone please

abrightwell (approved):
Sorry about that, the request got lost in the pile.

src/cb.cr
```diff
@@ -1,15 +1,15 @@
 module CB
   EID_PATTERN = /\A[a-z0-9]{25}[4aeimquy]\z/
 
-  VERSION       = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
-  BUILD_RELEASE = {{ flag?(:release) }}
-  BUILD_DATE    = {{ `date -u +"%Y%m%d%H%M"`.chomp.stringify }}
-  BUILD_ID      = begin
+  BUILD_RELEASE  = {{ flag?(:release) }}
+  SHARDS_VERSION = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
```
rbishop, line 5: Does this mean a user of the bridge CLI needs to have the crystal compiler/toolchain installed in order to run a precompiled bridge CLI?

fdr: No. Macro expansion at compile time.

rbishop: Ah I missed the curly braces around the shell invocation. That's pretty sweet!

PR #13, opened 2022-02-25
Add `cb upgrade` command.

abrightwell:
Adds the following commands for managing cluster upgrades:

* `cb upgrade start` starts an upgrade.
* `cb upgrade status` shows the status of cluster upgrades.
* `cb upgrade cancel` cancels cluster upgrades.

PR #14, opened 2022-03-09
Action Refactor

abrightwell:
As I've been adding support for new commands and actions, I found that it might be beneficial to start refactoring the previous implementations into actions as well. This is more of consistency in implementation that will hopefully support more advanced future features.

I have not addressed login in this changeset. As I believe that's a special case. ~Also, I have not updated the prior actions to utilize the new ClusterAction sub class. I was hoping to effect those changes in a future PR.~

~Unfortunately, includes changes also include in #13. I don't think it's possible to compare across forks against an existing PR. I'll rebase/update after #13 is approved/merged/etc.~

~Based on #15 until merged.~


will:
There is one change I've been meaning to do that I think will conflict a lot with this unfortunately. I'm putting it together now spurred on by this and I'll have to you take a look at it so we can discuss the pros/cons.

will:
Same thing with this. GH auto closed it because I merged my branch and I cannot reopen this.

abrightwell:
Hmm... it won't let me reopen it either. :thinking:

abrightwell:
Ok, looks like it auto closed due to the branch being deleted. I briefly restored, reopened, rebased, re-deleted.

PR #15, opened 2022-03-10
Add CLI input setter macros

will:
Use macros to define both the property and the initial format
transformation and validation for input that comes from the CLI
arguments. This unifies a lot of the copy and pasted logic, and also
adds some missed validations on eids that was easy to overlook.

@abrightwell This may (or may not) make some of the new subclassing you added in the other pr not as necessary. I've been meaning to add these macros for a while, but it was a super backburner idea. Sorry to only do this after you sent in the other PR.

### DIFF
--- a/src/cb/action.cr
+++ b/src/cb/action.cr
@@ -19,6 +19,53 @@ module CB
 
     abstract def run
 
+    macro eid_setter(property, description = nil)
+      property {{property}} : String?
+
+      def {{property}}=(str : String)
+        raise_arg_error {{description || property.stringify.gsub(/_/, " ")}}, str unless str =~ EID_PATTERN
+        @{{property}} = str
+      end
+    end
+
+    # For simple identifiers such as region names, or plan names where we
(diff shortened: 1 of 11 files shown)


abrightwell (approved):
Nice! Yeah, this definitely helps overall. If nothing else, it certainly simplifies #14 a great deal I think. I'll be revisiting that. :+1:

src/cb/action.cr
```diff
@@ -19,6 +19,53 @@ module CB
...
+      property {{property}} : Int32?
+
+      def {{property}}=(str : String)
+        self.{{property}} = str.to_i(base: 10, whitespace: true, underscore: true, prefix: false, strict: true, leading_zero_is_octal: false)
+      rescue ArgumentError
+        raise_arg_error {{property.stringify}}, str
+      end
+    end
+
+    # Not: unlike the other macros, this one does not create a nilable boolean,
```
abrightwell, line 52: `s/Not/Note`?

PR #16, opened 2022-03-10
Add `get_team` to `client`

abrightwell:
Previously this call was not supported. Now it is, so I've added it such that items like `print_team_slash_cluster` can leverage it, instead of having to iterate over the list of teams.

Based on #15 and #14 until merged.


will:
I just merged my branch and it auto closed this, I don’t know why

will:
I cannot reopen it

will:
Nice, I didn't notice that you could get an individual team now. That always made getting the team name cumbersome, so this is a lot nicer

PR #18, opened 2022-03-23
Add `cb role` command and other updates.

abrightwell:
This PR adds the `cb role` command to allow for managing cluster roles. As well, it updates the `cb uri` command to allow for a role name to be provided via the `--role` flag.

```
> ./bin/cb role --help
Usage: cb role <create|update|destroy>
    create                           Create new role for a cluster
    update                           Update a cluster role
    destroy                          Destroy a cluster role

> ./bin/cb role create --help
Usage: cb role create <--cluster>
    --cluster ID                     Choose cluster

> ./bin/cb role update --help
Usage: cb role update <--cluster> <--name> [--mode] [--rotate-password]
    --cluster ID                     Choose cluster
    --name NAME                      Role name
    --read-only <true|false>         Read-only
    --rotate-password <true|false>   Rotate password
    
> ./bin/cb role destroy --help
Usage: cb role destroy <--cluster> <--name>
    --cluster ID                     Choose cluster
    --name NAME                      Role name
```

```
> ./bin/cb uri --help
Usage: cb uri <cluster id> [--role]
    --role NAME                      Role name (default: default)
```

### DIFF
--- a/spec/cb/cluster_uri_spec.cr
+++ b/spec/cb/cluster_uri_spec.cr
@@ -1,0 +1,72 @@
+require "../spec_helper"
+
+private class ClusterURITestClient < CB::Client
+  ACCOUNT = Account.new(
+    id: "123",
+    name: "user",
+  )
+
+  ROLE = Role.new(
+    name: "u_" + ACCOUNT.id,
+    password: "secret",
+    uri: URI.parse "postgres://u_123:secret@localhost:5432/postgres",
+  )
+
+  def get_account
+    return ACCOUNT
+  end
+
+  property p_get_role : Proc(Role) = -> : Role { return ROLE }
+
+  def get_role(id, name)
+    return p_get_role.call
+  end
+end
+
+describe CB::ClusterURI do
+  it "ensures 'default' if role not specified" do
+    action = CB::ClusterURI.new(ClusterURITestClient.new(TEST_TOKEN))
+    action.output = IO::Memory.new
+
+    action.call
+
+    action.role_name.should eq "default"
+  end
+
+  it "#run errors on invalid role" do
+    action = CB::ClusterURI.new(ClusterURITestClient.new(TEST_TOKEN))
+    action.output = IO::Memory.new
+
+    action.role_name = "invalid"
+
+    msg = /invalid role: 'invalid'/
+
+    expect_cb_error(msg) { action.call }
+  end
+
+  it "#run handles client errors" do
+    c = ClusterURITestClient.new(TEST_TOKEN)
+    c.p_get_role = -> : CB::Client::Role {
+      raise CB::Client::Error.new("", "",
+        HTTP::Client::Response.new(HTTP::Status::BAD_REQUEST))
+    }
+
+    action = CB::ClusterURI.new(c)
+    action.output = output = IO::Memory.new
+
+    msg = /invalid input/
+
+    expect_cb_error(msg) { action.call }
+  end
+
+  it "#run prints uri" do
+    c = ClusterURITestClient.new(TEST_TOKEN)
+
+    action = CB::ClusterURI.new(c)
+    action.output = output = IO::Memory.new
+
+    action.call
+
+    output.to_s.should eq ClusterURITestClient::ROLE.uri.to_s
+  end
+end
--- a/src/cb/action.cr
+++ b/src/cb/action.cr
@@ -49,10 +49,27 @@ module CB
       end
     end
 
-    # Not: unlike the other macros, this one does not create a nilable boolean,
+    # Note: unlike the other macros, this one does not create a nilable boolean,
     # and instead creates one that defaults to false
     macro bool_setter(property)
       property {{property}} : Bool = false
+
+      def {{property}}=(str : String)
+        case str.downcase
+        when "true"
+          self.{{property}} = true
+        when "false"
+          self.{{property}} = false
+        else
+          raise_arg_error {{property.stringify}}, str
+        end
+      end
+    end
+
+    # Nilable boolean setter. This is useful for fields where nil can have a
+    # meaningful value beyond being falesy.
+    macro bool_setter?(property)
+      property {{property}} : Bool?
 
       def {{property}}=(str : String)
         case str.downcase
--- a/src/cb/client.cr
+++ b/src/cb/client.cr
@@ -28,12 +28,20 @@ class CB::Client
       end
     end
 
-    def unauthorized?
-      resp.status == HTTP::Status::UNAUTHORIZED
+    def bad_request?
+      resp.status == HTTP::Status::BAD_REQUEST
+    end
+
+    def forbidden?
+      resp.status == HTTP::Status::FORBIDDEN
     end
 
     def not_found?
       resp.status == HTTP::Status::NOT_FOUND
+    end
+
+    def unauthorized?
+      resp.status == HTTP::Status::UNAUTHORIZED
     end
   end
 
--- a/src/cb/client.cr
+++ b/src/cb/client.cr
@@ -169,13 +169,6 @@ class CB::Client
     raise Program::Error.new "cluster #{id.colorize.t_id} does not exist, or you do not have access to it"
   end
 
-  jrecord Role, name : String, password : String, uri : URI
-
-  def get_cluster_default_role(id)
-    resp = get "clusters/#{id}/roles/default"
-    Role.from_json resp.body
-  end
-
   # https://crunchybridgeapi.docs.apiary.io/#reference/0/clusters/post
   def create_cluster(cc)
     body = {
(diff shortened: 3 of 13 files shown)


src/cb/action.cr
```diff
@@ -66,6 +66,23 @@ module CB
       end
     end
 
+    # Nilable boolean setter. This is useful for fields where nil can have a
+    # meaningful value beyond being falesy.
+    macro bool_setter?(property)
```
abrightwell, line 71: I needed this so that we could definitively know when a particular flag was set and not simply omitted. For example, `--read-only` on `cb role update`. If we just defaulted to false then the it would update the role every time, possibly setting the 'mode' of 'flavor' of the role to be be `write`. :grimacing:

src/cb/client.cr
```diff
@@ -161,13 +169,6 @@ class CB::Client
     raise Program::Error.new "cluster #{id.colorize.t_id} does not exist, or you do not have access to it"
   end
 
-  jrecord Role, name : String, password : String, uri : URI
-
-  def get_cluster_default_role(id)
-    resp = get "clusters/#{id}/roles/default"
-    Role.from_json resp.body
-  end
-
```
abrightwell, line 170: Moved down below, where the rest of the role related methods are defined.

src/cb/psql.cr
```diff
@@ -6,12 +6,13 @@ class CB::Psql < CB::Action
 
   def run
     c = client.get_cluster cluster_id
-    uri = client.get_cluster_default_role(cluster_id).uri
+    uri = client.get_role(cluster_id, "default").uri
+    raise Error.new "null uri" if uri.nil?
```
abrightwell, line 10: Because we had to make URI nilable, we now have to check for that here (and in scope). Perhaps this could be handled cleaner?

src/cb/psql.cr
```diff
@@ -63,16 +64,4 @@ class CB::Psql < CB::Action
...
-    team_name = team_name_for_cluster c
-    io << team_name << "/" if team_name
-    io << c.name.colorize.t_name << "\n"
-    team_name
-  end
-
-  private def team_name_for_cluster(c)
-    # no way to look up a single team yet
-    client.get_teams.find { |t| t.id == c.team_id }.try &.name.colorize.t_alt
-  end
```
abrightwell, line 77: Missed this in the original refactor to pull these functions up into `CB::Action`

src/cb/role.cr
```diff
@@ -0,0 +1,76 @@
+require "./action"
+
+module CB
+  # Valid cluster role names.
+  VALID_CLUSTER_ROLES = ["application", "default", "postgres", "user"]
```
abrightwell, line 5: Figured I'd make these top-level to the module as opposed to defined on the `RoleAction` class. Seems reasonable, I think, that they should be considered global to all `CB` related items as it's not just this class that owns them.

src/stdlib_ext.cr
```diff
@@ -34,8 +34,10 @@ class IO
 end
 
 macro jrecord(name, *properties)
+  @[JSON::Serializable::Options(emit_nulls: true)]
```
abrightwell, line 37: So, without this `uri` would not serialize correctly. I'm not sure this is the best approach as it obviously applies to all `jrecord` expansions which might not be desired. However, I'm not sure I see an immediate downside to this at the moment.

src/cb/client.cr
```diff
@@ -291,6 +292,36 @@ class CB::Client
...
+  end
+
+  # https://crunchybridgeapi.docs.apiary.io/#reference/0/clustersclusteridrolesrolename/get-role
+  def get_role(cluster_id, role_name)
+    resp = get "clusters/#{cluster_id}/roles/#{role_name}"
+    Role.from_json resp.body
+  end
+
+  # https://crunchybridgeapi.docs.apiary.io/#reference/0/clustersclusteridrolesrolename/update-role
+  def update_role(cluster_id, role_name, ur)
```
brandur, line 314: OOC, what's "ur"? Might be better to call this a body or request body if that's what it is.

brandur (approved):
Seems good from my perspective!

One question. These usage docs that are generated:

```
> ./bin/cb uri --help
Usage: cb uri <cluster id> [--role]
    --role NAME                      Role name (default: default)
```

Would it be possible to eventually have some sort of docstring on there describing what the command does? Some are reasonably obvious, but IMO some like this aren't — "uri" is a pretty generic term, and I'd have to guess that what "cb uri" does is give you a Postgres connection string.

src/cb/client.cr
abrightwell: heh... yeah... `ur` -> `updated role`... I was just following other similar methods for consistency. But yeah, I definitely agree it could be more clear on what it is.

abrightwell:
> Seems good from my perspective!
> 
> One question. These usage docs that are generated:
> 
> ```
> > ./bin/cb uri --help
> Usage: cb uri <cluster id> [--role]
>     -h, --help                       Show this help
>     --role NAME                      Role name (default: default)
> ```
> 
> Would it be possible to eventually have some sort of docstring on there describing what the command does? Some are reasonably obvious, but IMO some like this aren't — "uri" is a pretty generic term, and I'd have to guess that what "cb uri" does is give you a Postgres connection string.

Yeah, I think there can/should be some improvements around this. I was also looking into the potential behind having some `man` pages we could generate for it as well to bring more details to all of it as well.

Also, another improvement I want to make is around the flag descriptions themselves. Would be great to have more detail there in some way.

spec/cb/cluster_uri_spec.cr
```diff
@@ -0,0 +1,72 @@
...
+
+  def get_account
+    return ACCOUNT
+  end
+
+  property p_get_role : Proc(Role) = -> : Role { return ROLE }
+
+  def get_role(id, name)
+    return p_get_role.call
+  end
```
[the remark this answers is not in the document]
will: I think it's good yeah. As long as we keep the same interface between the client we give things in the test, and the one that is used for real, it's okay to have the test one have more hooks to configure things. And by subclassing like this the compile type class inferred should be `Client+` iirc, and that will make sure all the types are good and the real code isn't relying on anything that is only in the tests

src/cb/psql.cr
will: Yeah, we could add a `get_uri` method maybe that wraps up checking that it's not nil, returning an error message if it is (and we really expect it to not be), and always returns a non-nilable uri

src/cb/role.cr
will, line 5:
```suggestion
  VALID_CLUSTER_ROLES = Set{"application", "default", "postgres", "user"}
```

This is a very small thing, I probably wouldn't have noticed except you called it out with the other comment :)

src/stdlib_ext.cr
will: Yeah. If we do start seeing problems with the extra nulls in the serialized things, we can put it back, and then not use `jrecord` just for that one. Or add an option to the jrecord macro or something

src/cb/action.cr
will: I think we define one of these in terms of the other?, or have one that takes an option. But it's fiddly stuff and I can take a look after it's merged.

will (approved):
Just the one small thing I think. I can take care of it tomorrow if you don’t get to it first.

spec/cb/cluster_uri_spec.cr
```diff
@@ -0,0 +1,72 @@
...
+
+    expect_cb_error(msg) { action.call }
+  end
+
+  it "#run handles client errors" do
+    c = ClusterURITestClient.new(TEST_TOKEN)
+    c.p_get_role = -> : CB::Client::Role {
+      raise CB::Client::Error.new("", "",
+        HTTP::Client::Response.new(HTTP::Status::BAD_REQUEST))
+    }
```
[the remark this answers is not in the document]
will: Yeah we could take a look at patterns in our usage of them in the test and make more helper methods. Idk if we have enough now to figure out what'd be useful, or if it'd make sense to wait for a few more to accumulate

src/stdlib_ext.cr
will: I think this broke provisioning

```
fdr@shrike ~/c/owl (redo-acl)> cb version
cb v1.2.0 (202204071829)
fdr@shrike ~/c/owl (redo-acl)> cb create --plan standard-0 -p aws -r us-west-2 -t qpfaodoqxvac7j65v5ayehzthe
error: 400 Bad Request
       POST to /clusters
       message: Invalid JSON in request body: json: unknown field "ha".
       request_id: 04a37e55-75a8-4087-b9b9-6ec8e1da39ae
```

unless platform recently added a thing to reject it, and it didn't when this came in?

will: cc @abrightwell @brandur did platform between now and march 29th start rejecting unknown nulls?

brandur: @will Yeah, probably this: https://crunchydata.slack.com/archives/C012GUR4XAT/p1650473167444349

It was done with the premise to help raise bugs like this one — should be `is_ha` I think.

will: It was changed from `ha` to `is_ha` here https://github.com/CrunchyData/bridge-cli/commit/f7db90539907e70707270fce013fee54868ff843 but it seems the `ha` one came back in some merge error or something. The github blame see previous version isn't correctly showing where it came back from, so haven't been able to figure it out yet.

brandur: Hmmm, weird ...

PR #19, opened 2022-03-31
Add Team Management Commands

abrightwell:
Adds commands for managing teams.

Specifically, we add the following commands:
    * `create` - create a new team with the provided name.
    * `list` - list the teams that are available to the requesting user.
    * `info` - show detailed information about a team.
    * `update` - update the billing email, name or sso enforcement of a team.
    * `destroy` - delete a team.
    
 ```
 > cb team --help
Usage:
    cb team <command>

Available Commands:
    create                           Create a new team.
    destroy                          Delete a team.
    info                             Show a teams details.
    list                             List available teams.
    update                           Update a team.

Options:
```

Deprecation Example:

```
> cb teams --help
Deprecated: Prefer use of 'cb team list' instead.

Usage:
    cb teams

Options:
```

### CODE REVIEW

File: src/cli.cr [hunk 1]
```diff
@@ -14,6 +14,10 @@ macro set_action(cl)
   action = CB::{{cl}}.new PROG.client
 end
 
+def show_deprecated(msg : String)
+  printf "%s: %s\n\n", "Deprecated".colorize.t_warn, msg
+end
```
abrightwell, line 19: I had initially thought about maybe adding a 'deprecated' property to the `Action`. But for now, this seemed simplest given that it's just one we are currently handling.

src/cli.cr
```diff
@@ -29,11 +33,42 @@ Raven.configure do |config|
 end
 
 class OptionParser
-  # for hiding an option from help, omit description
+  # Allows for a hiding an option from help.
   def on(flag : String, &block : String ->)
     flag, value_type = parse_flag_definition(flag)
     @handlers[flag] = Handler.new(value_type, block)
   end
+
```
abrightwell, line 41: Here I decided to add some more organization to the help message for commands.

Specifically, I wanted to give explicit separation of commands and flags... as well as having some sorted ordering to them as well to ensure consistency of experience across all commands.

I had also thought about adding the `Deprecated` message here if set on an `Action`, however, since I opted to go with the above for now. I left it out. Also, one thing that this would only show such a message on the `--help` for the command. Which I think should also be presented if/when the command is also executed... in some fashion.

brandur (approved):
Will probably have to rely on Will for code review, but the CLI design looks solid to me.

src/cli.cr
```diff
@@ -29,11 +33,42 @@ Raven.configure do |config|
...
+      io << commands.sort.join io, "\n"
+      io << "\n"
+    end
+
+    if !flags.empty?
+      io << "\n" << "Options".colorize.bold << ":\n"
+      flags.sort! { |a, b| a.lstrip(" -") <=> b.lstrip(" -") }
+      flags.join io, '\n'
+    end
+  end
```
will, line 71: These are all very minor things 

```suggestion
    unless commands.empty?
      io << '\n' << "Available Commands".colorize.bold << ":\n"
      io << commands.sort.join io, '\n'
      io << '\n'
    end

    unless flags.empty?
      io << '\n' << "Options".colorize.bold << ":\n"
      flags.sort.join io, '\n'
    end
  end
```

`unless` - as long as it's a very simple conditional, prefer `unless` to `if !`. Though once you get any sort of `||` or `&&` or an `else`, go back to `if`
single quote chars - if it is just one char, using single quotes gives you a `Char` instead of a `String`. This might even get optimized out anyway by llvm, so might not even matter at all, and then also it's a very small thing even if it did matter. But you had it both ways, might as well just use the char version
no `sort!` - this mutates the array, but it's just going to be thrown out anyway, can just chain it to the `join`, so you can just use `sort`.   Then also I went even further and removed the block. I think it should still sort the same since youre doing the same thing to both sides. I might be missing some detail though



``` crystal
a # => ["-flag1", "-aflag", "-zflag", "-okflags"]
a.sort # => ["-aflag", "-flag1", "-okflags", "-zflag"]
a.sort do |a, b|
  (a.lstrip(" -")) <=> (b.lstrip(" -"))
end # => ["-aflag", "-flag1", "-okflags", "-zflag"]
```

will (changes requested):
Just a few small things, nothing at all on the new command parts, that's all good. But also now that we have a good amount of customization to the option parser, maybe it should be in it's own file?

src/cli.cr
abrightwell: > unless - as long as it's a very simple conditional, prefer unless to if !. Though once you get any sort of || or && or an else, go back to if

:+1: 

> single quote chars - if it is just one char, using single quotes gives you a Char instead of a String. This might even get optimized out anyway by llvm, so might not even matter at all, and then also it's a very small thing even if it did matter. But you had it both ways, might as well just use the char version

:+1:

abrightwell: > But also now that we have a good amount of customization to the option parser, maybe it should be in it's own file?
:+1:

will: Ah, yeah if it was to make one and two `-` not matter then you're right.

One thing if both sides are the same, you can use `sort_by` instead of `sort`. `a.sort_by(&.lstrip(" -"))`

abrightwell: Oh, I like that MUCH better. :+1:

abrightwell:
All requested changes have been incorporated. :+1:

PR #20, opened 2022-04-06
add cb rename

will:
I noticed you couldn't rename replicas in the web ui, but it does work in the API (which it should) and that cb didn't have it yet.

Also the `state` field comes down on get but not patch? I don’t get why that is, so I made it nilable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `status` sub-commands.
 - `cb detach` command added to detach clusters.
 - `cb restart` command added to restart clusters.
+- `cb rename` command added to rename clusters.
 
 ## [1.1.0] - 2022-01-27
 ### Added
--- a/spec/cb/completion_spec.cr
+++ b/spec/cb/completion_spec.cr
@@ -78,6 +78,11 @@ describe CB::Completion do
     result.should eq ["abc\tmy team/my cluster"]
   end
 
+  it "rename suggests clusters" do
+    result = parse("cb rename ")
+    result.should eq ["abc\tmy team/my cluster"]
+  end
+
   it "uri suggests clusters" do
     result = parse("cb uri ")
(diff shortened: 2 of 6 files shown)


abrightwell:
> Also the state field comes down on get but not patch? I don’t get why that is, so I made it nilable.

Digging a little, I believe I see the issue with that. It would appear to be related to Owl returning `204 No Content` on this request. I could get a PR for both API and Owl pretty quickly, if that's not the desired behavior. I want to say it isn't anymore as I think we've been incrementally moving away from it on other endpoints.

@brandur thoughts?

brandur:
@abrightwell Yeah, definitely true we're moving away from 204s in general. Most of them are gone already and I think that's one of the few that's left.

That said, `state` in particular is a bit of an odd one. Basically, Platform doesn't have the information, which is why it's not returned on on "cluster update". On "cluster get", we go to Owl to get the information, but it's one of two fields we need Owl for, and IMO we might want to just move `state` to the `/status` endpoint instead:

https://linear.app/crunchydata/issue/BR-1076/cluster-get-endpoint-try-to-remove-oldest-backup-at-state

But it's a bit scary because we have public consumers on this endpoint now, and I'm afraid we might break someone. I think I asked Craig a while back to try and reach out to some of the customers we know to be using the endpoint and seeing if they depend on `state`, but I'm not sure it happened.

We could definitely add it to for "cluster update" to make it consistent with "cluster get", but it'd still be missing from "cluster list", which doesn't go to Owl at all, and therefore still be nilable :)

So to sum up, unfortunately it's going to be tough to make state completely non-nilable for now. A couple things that we could start right away:

* Change that Owl response from 204 to 200 — can't hurt anyway and would make Owl's endpoints more normalized.
* More investigation on whether we can safely drop `state` out of cluster response.

will:
It's okay nilable to me

abrightwell:
Yeah it doesn't affect anything so certainly shouldn't hold anything up.

abrightwell (approved):

PR #21, opened 2022-04-12
Add Team Member Management Commands

abrightwell:
Add support for `team-member` management.

Adds command support for `team-member` management.

Implements:

`add` - add member to a team.
`info` - show details of a team member.
`list` - list current members of a team.
`update` - updates member of team - currently only supports updating role.
`remove` - removes member from team.

```
> cb team-member --help
Usage:
    cb team-member <command>

Available Commands:
    add                              Add a team member
    info                             Show team member details
    list                             List members of a team
    remove                           Remove a member from a team
    update                           Update a member of a team

Options:
```

### DIFF
--- a/spec/cb/team_member_spec.cr
+++ b/spec/cb/team_member_spec.cr
@@ -1,0 +1,219 @@
+require "../spec_helper"
+
+private class TeamMemberTestClient < CB::Client
+  MEMBERS = [
+    CB::Client::TeamMember.new(
+      id: "abc",
+      team_id: "pkdpq6yynjgjbps4otxd7il2u4",
+      account_id: "4pfqoxothfagnfdryk2og7noei",
+      role: "member",
+      email: "test@example.com",
+    ),
+  ]
+
+  def get_team(team_id)
+    return Team.new(
+      id: "pkdpq6yynjgjbps4otxd7il2u4",
+      name: "Test Team",
+      is_personal: false,
+      role: nil,
+      billing_email: nil,
+      enforce_sso: nil,
+    )
+  end
+
+  def create_team_member(team_id, params : TeamMemberCreateParams)
+    return TeamMember.new(
+      id: "",
+      team_id: team_id.to_s,
+      account_id: "abc123",
+      role: params.role,
+      email: params.email,
+    )
+  end
+
+  def get_team_member(team_id, account_id)
+    return MEMBERS[0]
+  end
+
+  def list_team_members(team_id)
+    return MEMBERS
+  end
+
+  def update_team_member(team_id, account_id, role)
+    return MEMBERS[0]
+  end
+
+  def remove_team_member(team_id, account_id)
+    return MEMBERS[0]
+  end
+end
+
+describe CB::TeamMemberAdd do
+  it "validates that required arguments are present" do
+    action = CB::TeamMemberAdd.new(TeamMemberTestClient.new(TEST_TOKEN))
+
+    msg = /Missing required argument/
+    expect_cb_error(msg) { action.validate }
+
+    action.team_id = "pkdpq6yynjgjbps4otxd7il2u4"
+    action.email = "test@example.com"
+    action.validate.should eq true
+
+    action.role = "some role"
+    msg = /invalid role '#{action.role}'/
+    expect_cb_error(msg) { action.validate }
+  end
+
+  it "#run prints confirmation" do
+    action = CB::TeamMemberAdd.new(TeamMemberTestClient.new(TEST_TOKEN))
+    action.output = output = IO::Memory.new
+
+    action.team_id = "pkdpq6yynjgjbps4otxd7il2u4"
+    action.email = "test@example.com"
+
+    action.call
+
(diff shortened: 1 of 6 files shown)


spec/cb/team_member_spec.cr
```diff
@@ -0,0 +1,219 @@
...
+
+describe CB::TeamMemberList do
+  it "validates that required arguments are present" do
+    action = CB::TeamMemberList.new(TeamMemberTestClient.new(TEST_TOKEN))
+
+    msg = /Missing required argument/
+    expect_cb_error(msg) { action.call }
+  end
+
+  it "#run" do
```
brandur, line 111: May be a pre-existing problem, but generally these spec names are called `it` so that they read as a sentence like "it succeeds" or "it validates that required arguments are present". For method names, you'd normally put those in a `describe` block with an `it` test case wrapped below. Not a big deal though.

src/cb/completion.cr
```diff
@@ -509,6 +512,116 @@ class CB::Completion
...
+      return teams
+    end
+
+    if last_arg? "--account", "--email"
+      suggest_none
+    end
+
+    suggest = [] of String
+    suggest << "--team\tteam ID" unless has_full_flag? :team
+    suggest << "--account\tuser's account ID" unless has_full_flag?(:account) || has_full_flag?(:email)
```
brandur, line 573: Another pre-existing one, but most Rubyists (I thinkk I'm fair in saying that) consider it bad practice to use complex `unless` cases because they're really hard to mentally parse.

My last job banned `unless` that had more one condition via Rubocop.

This one isn't too bad, but worth thinking about.

brandur (commented):
Seems good to me, but like last one will leave full review to Will.

spec/cb/team_member_spec.cr
will: Part of it I think is per-test setup is a little more cumbersome in the stdlib crystal test than ruby's 3rd party rspec, so you end up making more combined test that have more than one assertion. 

However I recently came across https://gitlab.com/arctic-fox/spectator/-/wikis/Mocks-and-Doubles which looks pretty neat, and might be worth switching to.

src/cb/client.cr
```diff
@@ -119,6 +119,10 @@ class CB::Client
     def name
       is_personal ? "personal" : @name
     end
+
+    def to_s : String
+      return sprintf("%s (%s)", id.colorize.t_id, name.colorize.t_name)
+    end
```
will, line 125:
```suggestion
    def to_s(io : IO)
      io << id.colorize.t_id << " (" << name.colorize.t_name << ")"
    end
```

https://crystal-lang.org/reference/1.4/guides/performance.html#dont-create-intermediate-strings-when-writing-to-an-io explains why better than I could

src/cb/completion.cr
will: I agree with all of this. In general once a conditional gets any more complicated than a single thing, you should avoid `unless`.

However for this particular set of lines `unless` is very good. "Add this option unless we have that flag" just is really great. And going to "add this option unless there is that flag or that other flag" isn't too much of a stretch to that.

Also switching it to an `if` in this case I think makes it worse than the `unless` (which is often not the case)

```crystal
suggest << "--account\tuser's account ID" if !(has_full_flag?(:account) || has_full_flag?(:email))
# or
suggest << "--account\tuser's account ID" if !has_full_flag?(:account) && !has_full_flag?(:email)
 ```

So for this particular instance keeping the `unless` is the best thing I think. But for sure it's a keyword to pay attention to abusing.

File: src/cb/client.cr (hunk 3 above)
Comment (abrightwell) [reply src/cb/client.cr:125]: Ah! Very cool. I didn't realize that's how it was handling things under the hood, but it makes perfect sense. :+1:

File: src/cb/completion.cr (hunk 2 above)
Comment (will) [reply src/cb/completion.cr:573]: `has_full_flag?` can take multiple flags, but it wants *all* flags to be there for it to be true.

We could maybe make a method 

```crystal
def has_any_full_flag?(*names : Symbol)
   names.any? { |n| @full_flags.includes? n }
end

# then it could be
<< "thing" unless has_any_full_flag? :a, :b

```

Which might be a bit easier to reason about? Not fully sure either way.

src/cb/team_member.cr
```diff
@@ -0,0 +1,184 @@
...
+    end
+
+    raise Error.new "Must only use '--account' or '--email' but not both." if account_id && email
+  end
+
+  private def get_member_by_email(team_id, email)
+    members = client.list_team_members(team_id)
+    members.find { |tm| tm.email == email }
+    # return tm unless tm.nil?
+    # raise Error.new "Unknown team member '#{email.colorize.t_name}'"
```
will, line 38: Was this just something you were experimenting with? Should we have it in for real?

will (changes requested):
Just really the to_s thing, then if you want to explore the .any? option of the full_flag thing, but just if that feels nicer

src/cb/team_member.cr
abrightwell: Yeah, meant to remove it... removed now though. :+1:

PR #22, opened 2022-04-19
Ameba

will:
This is a style enforcing linter like standardrb, etc. I don’t have that strong opinions on this either way, but the last PR there was some style discussion and it might be easy to defer to this?

### DIFF
--- a/src/cb/cluster_upgrade.cr
+++ b/src/cb/cluster_upgrade.cr
@@ -31,7 +31,7 @@ class CB::UpgradeStart < CB::Upgrade
       response = input.gets
 
       if c.name == response
-        confirmed = true
+        self.confirmed = true
       else
         raise Error.new "Response did not match, did not upgrade the cluster"
       end
--- a/src/cb/restart.cr
+++ b/src/cb/restart.cr
@@ -15,7 +15,7 @@ class CB::Restart < CB::Action
       response = input.gets
 
       if c.name == response
-        confirmed = true
+        self.confirmed = true
       else
         raise Error.new "Response did not match, did not restart the cluster"
       end
--- a/src/cb/scope.cr
+++ b/src/cb/scope.cr
@@ -29,7 +29,7 @@ class CB::Scope < CB::Action
       raise Error.new("unknown suite '#{suite.inspect}'")
     end
 
-    to_run = checks.uniq.sort_by(&.name)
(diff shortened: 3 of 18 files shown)


src/cb/cluster_upgrade.cr
```diff
@@ -31,7 +31,7 @@ class CB::UpgradeStart < CB::Upgrade
       response = input.gets
 
       if c.name == response
-        confirmed = true
+        self.confirmed = true
```
will, line 34: This was a proper find, before it was just setting a local variable

src/cb/restart.cr
```diff
@@ -15,7 +15,7 @@ class CB::Restart < CB::Action
       response = input.gets
 
       if c.name == response
-        confirmed = true
+        self.confirmed = true
```
will, line 18: This too

src/cb/scope.cr
```diff
@@ -29,7 +29,7 @@ class CB::Scope < CB::Action
       raise Error.new("unknown suite '#{suite.inspect}'")
     end
 
-    to_run = checks.uniq.sort_by(&.name)
+    to_run = checks.uniq.sort_by!(&.name)
```
will, line 32: I didn't know this one 

```
src/cb/scope.cr:32:26 [Correctable]
[C] Performance/ChainedCallWithNoBang: Use bang method variant `sort_by!` after chained `uniq` call
> to_run = checks.uniq.sort_by(&.name)
                       ^-----^
```

```
/tmp> cat a.cr
require "benchmark"

arr = [2,4,-5,2]

Benchmark.ips do |x|
  x.report("orig") { arr.dup.uniq.sort_by(&.abs)}
  x.report("bang") { arr.dup.uniq.sort_by!(&.abs)}
end
/tmp> crystal run --release a.cr
orig   7.04M (141.99ns) (± 2.24%)  224B/op   1.19× slower
bang   8.39M (119.19ns) (± 5.55%)  176B/op        fastest
```

uses less memory and is faster. It makes sense because before it'd make an array for the uniq then yet another array for the sort, and with ! it just mutates the one it made for uniq. But that did not occur to me.

abrightwell:
WFM. Gotta start somewhere. :+1:

abrightwell (approved):

PR #24, opened 2022-05-04
Update `restart` command to support service restart.

abrightwell:
The API recently added the ability to restart both the postgres service
and the full server. Here we're adding the ability to specify which to
restart. By default, the `restart` command will restart the postgres
service. However, using the `--full` flag will perform a full restart of
the server.

### DIFF
--- a/src/cb/completion.cr
+++ b/src/cb/completion.cr
@@ -48,14 +48,16 @@ class CB::Completion
         single_cluster_suggestion
       when "create"
         create
+      when "detach"
+        detach
       when "firewall"
         firewall
       when "logdest"
         logdest
       when "psql"
         psql
-      when "restart", "detach"
-        restart_or_detach
+      when "restart"
+        restart
       when "role"
         role
       when "team", "teams"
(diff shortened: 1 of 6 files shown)


src/cli.cr
```diff
@@ -194,11 +194,10 @@ op = OptionParser.new do |parser|
...
   parser.on("restart", "Restart a cluster") do
     restart = set_action Restart
-    parser.banner = "cb restart <cluster id> [--confirm]"
+    parser.banner = "cb restart <cluster id> [--confirm] [--full]"
 
-    parser.on("--confirm", "Confirm cluster restart") do
-      restart.confirmed = true
-    end
+    parser.on("--confirm", "Confirm cluster restart") { restart.confirmed = true }
+    parser.on("--full <true|false>", "Full restart of server (default: false)") { |arg| restart.full = arg }
```
abrightwell, line 200: I'm kinda on the fence here... feel like it could just be an arg-less option... :shrug: I'm cool with whichever seems most appropriate.

will: Yeah I think the arg-less version is probably better, let's switch to that

src/cb/completion.cr
```diff
@@ -403,10 +405,24 @@ class CB::Completion
...
+  def restart
+    return cluster_suggestions if @args.size == 2
+
+    if last_arg?("--confirm")
+      [] of String
+    end
+
+    if last_arg?("--full")
+      [] of String
+    end
```
will, line 417: Oh just noticed these the second time around. I don’t think they do anything, just create an array then it goes out of scope and GCd as is now. You probably meant to return the empty array, but the tests look right, and I think if you did return them, then you wouldn't get the completion for the other if you have one. So maybe these shouldn't be there at all?

abrightwell: Good catch. Fixed. :+1:

PR #26, opened 2022-05-10
Add ability to set different domains for logs subcommand

gitstashpop:
Using `CB_CLUSTER_DOMAIN`, you can specify an alternative domain
for the logs subcommand.

### DIFF
--- a/src/cb/logs.cr
+++ b/src/cb/logs.cr
@@ -6,11 +6,12 @@ require "ssh2"
 module CB
   class Logs < Action
     eid_setter cluster_id
+    property domain : String?
 
     def run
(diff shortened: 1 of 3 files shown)


abrightwell:
OOC, what's the use case here?

gitstashpop:
@abrightwell Running cb against staging. I was able to test `cb logs` on staging.

will (changes requested):
Thanks for sending this in, but I don’t like having to set two env vars to use staging :/ . I think we can have it where it either always looks up the proper hostnamme with the get cluster info api call. Or only do that if the cb_host env var is set if we want to save the one extra api call.

PR #27, opened 2022-05-11
Fix `unknown ca` issue.

abrightwell:
One common issue that has come up is an `unknown ca` error with `cb
psql` when the user has `~/.postgresql/postgresql.crt` defined. ~Since
this is behavior that is internal to `psql` there really isn't much `cb`
can do to work around it. Therefore, it seems prudent to at the very
least provide some documentation or guidance on how to get around this
issue as the error message from `psql` is not very intuitive as to the
cause.~

~So, here we add a `Troubleshooting` section to the `README` file to
capture this in order to assist users that might encounter this issue
moving forward.~

As a solution, we're simply overriding the `libpq` defaults but providing a
value for the `PGSSLCERT` and `PGSSLKEY` ENVs which will be
ignored by `libpq`.

### DIFF
--- a/src/cb/psql.cr
+++ b/src/cb/psql.cr
@@ -56,7 +56,7 @@ class CB::Psql < CB::Action
       s << "%[%033[36m%]#{c.name}%[%033m%]"
     end
 
-    psqlrc = File.tempfile(c.id, "pslrc")
+    psqlrc = File.tempfile(c.id, "psqlrc")
     File.copy("~/.psqlrc", psqlrc.path) if File.exists?("~/.psqlrc")
     File.open(psqlrc.path, "a") do |f|
       f.puts "\\set ON_ERROR_ROLLBACK interactive"
--- a/src/cb/psql.cr
+++ b/src/cb/psql.cr
@@ -26,6 +26,8 @@ class CB::Psql < CB::Action
       "PGDATABASE"    => uri.path.lchop('/'),
       "PGPORT"        => uri.port.to_s,
       "PSQLRC"        => psqlrc_path,
+      "PGSSLCERT"     => "dontuse",
+      "PGSSLKEY"      => "dontuse",
       "PGSSLMODE"     => "verify-ca",
       "PGSSLROOTCERT" => cert_path,
     })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the certificate file.
 
 ### Fixed
+- Fix `unknown ca` error with `cb psql`.
 - Fix cluster creation error after the API became stricter.
 
 ## [1.2.0] - 2022-04-07


src/cb/psql.cr
```diff
@@ -56,7 +56,7 @@ class CB::Psql < CB::Action
       s << "%[%033[36m%]#{c.name}%[%033m%]"
     end
 
-    psqlrc = File.tempfile(c.id, "pslrc")
+    psqlrc = File.tempfile(c.id, "psqlrc")
```
abrightwell, line 59: Unrelated to the troubleshooting docs, but stumbled on this while doing some digging on this particular issue. Captured as separate commit.

abrightwell:
Also, another thought that I had around this issue was to include maybe a `Warning` message if the files exist. Which might be something desired regardless. However, I figured this was at least a good start.

Here's what I was playing around with initially:

```
Warning: /home/abrightwell/.postgresql/postgresql.crt exists.
This may cause a verification error. If this occurs, then try renaming or removing this file and re-run this command.

connecting to personal/test-cluster-1
psql (14.2 (Ubuntu 14.2-1.pgdg18.04+1), server 13.6)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

personal/test-cluster-1/postgres=#
```

:shrug:

gitstashpop:
I think it's good to warn the user. Imo if we do that, we can unset sslcert/key/root that psql picks up from ~/.postgresql using the `sslcert= sslkey= sslroot=` change discussed in slack. Afaik, you can not connect to bridge using your own certificates, so it makes sense to never let users provide their own defaults.

abrightwell:
> I think it's good to warn the user. Imo if we do that, we can unset sslcert/key/root that psql picks up from ~/.postgresql using the `sslcert= sslkey= sslroot=` change discussed in slack. Afaik, you can not connect to bridge using your own certificates, so it makes sense to never let users provide their own certs or rely on the psql defaults.

Yeah, the other option not mentioned was to simply set the `PGSSLCERT` ENV to something random. That would work as well, but I wasn't entirely sure that's what we would want. Though, it is a simple solution that would mitigate the need for ALL of the above.

will:
I think setting all of the env vars to something (not random but something like `"dontuse"` or along those lines) could potentially be an ok solution. We are getting some default libpq behavior, but we know we don’t want those defaults in our case, so setting the env vars to something to stop the defaults makes sense to me. I'm not fully self-convinced by this argument, but it seems to make sense.

abrightwell:
> I think setting all of the env vars to something (not random but something like `"dontuse"` or along those lines) could potentially be an ok solution. We are getting some default libpq behavior, but we know we don’t want those defaults in our case, so setting the env vars to something to stop the defaults makes sense to me. I'm not fully self-convinced by this argument, but it seems to make sense.

Updated to this.

will:
Looks good, can you rebase then add a changelog entry please? Probably with the `### Fixed` subheading I think

PR #31, opened 2022-05-18
Add backup list

will:
Trying out a thing where the relevant methods on client are done in the same file as the action. It may be nice for things that are only ever used in one action to have it all in one place. If not it's easy enough to move it back into the client file later on I think

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -387,6 +387,19 @@ op = OptionParser.new do |parser|
     end
   end
 
+  #
+  # Backup Management
+  #
+  parser.on("backup", "Manage backups") do
+    parser.banner = "cb backup <list>"
+
+    parser.on("list", "List backups for a cluster") do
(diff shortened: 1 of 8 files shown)


src/cli.cr
```diff
@@ -387,6 +387,19 @@ op = OptionParser.new do |parser|
...
 
+  #
+  # Backup Management
+  #
+  parser.on("backup", "Manage backups") do
+    parser.banner = "cb backup <list>"
+
+    parser.on("list", "List backups for a cluster") do
+      list = set_action BackupList
+      parser.banner = "cb backup cpature <cluster id>"
```
abrightwell, line 398:
```suggestion
      parser.banner = "cb backup list <cluster id>"
```

abrightwell (approved):

will:
Going to wait on merging until all the apis are in

PR #34, opened 2022-05-19
Update `cb backups list`.

abrightwell:
Brings in line with final API response shape.

Suggested updates to #31 to resolve compatibility with API.


abrightwell:
@will these are just the changes I made locally to get this working. Wasn't sure the best way to 'suggest' them on #31. So figured I'd just create a PR against that PR's branch. :shrug: At any rate, these work with my testing.

PR #35, opened 2022-05-26
Add `backup token` support.

abrightwell:


src/cb/backup.cr
```diff
@@ -62,4 +86,42 @@ module CB
...
+      token = client.backup_token cluster_id
+      raise Error.new("backup token not created") if token.nil?
+
+      cred = if !token.aws.nil?
+               token.aws
+             elsif !token.azure.nil?
+               token.azure
+             else
+               "No Credentials"
+             end
```
abrightwell, line 107: This was kind of weird... I think. But figured it could get straightened out post-review if needed.

fdr: In the new version you could dispatch on "type" which is "s3" or "azure". You might object, saying: "hey! one's a family of products, one is a product" and I'd say "well, that's the literal you need in pgbackrest to configure it, here's a buck to pass"

fdr: (that said...it is probably slower than checking for nil pointers, and I guess what's the point?)

abrightwell: So, that's initially what I was trying to do. However, what I kept running up into was a nilable type error at compile time.

For instance, I had done something like:

```
case token.type
when "s3"
...
when "azure"
...
end
```
But I when I tried to reference `token.aws` below that it would bark at me due to the nilable aspect of that particular property. :shrug: 

At any rate, I'm definitely open to better ways around it. But thought maybe this is a 'good enough' for now solution.

Personally, I'm not terribly worried about it as I hope to come back in future refactors across all actions to hopefully make displaying/formatting output a little cleaner. Which I kind of feel like this at least tangentially relates.

cbandy: In Ruby, you can `case` without an expression, if that reads more clearly to you.

```ruby
      cred = case
             when !token.aws.nil?
               token.aws
             when !token.azure.nil?
               token.azure
             else
               "No Credentials"
             end
```

abrightwell: Yeah, you can do the same here.  Might go with that as seems 'cleaner'... ish.

PR #36, opened 2022-05-26
Add some error handling to `cb logs`.

abrightwell:
It appears that no exception handling was being done by this action.
Therefore it would output stacktraces to the user. While this doesn't
necessarily make the cause of the error easier to determine for the
user. It is at least a better look.  It would be great to tie in some
reporting on some of these, but that's for a potential future update.

### DIFF
--- a/src/cb/logs.cr
+++ b/src/cb/logs.cr
@@ -23,6 +23,8 @@ module CB
       while (read_bytes = ch.read(buffer.to_slice)) > 0
         output.write buffer.to_slice[0, read_bytes]
       end
+    rescue e
+      raise Program::Error.new(cause: e)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


will:
You can create exceptions that wrap other ones with `cause:`. That might be handy here? 

The one issue I found playing around with that is that the message doesn't propagate . However since we're mostly using our own exception type for doing things, we can fix that for ourselves by overriding `#message` 

```crystal
class CB::Error < Exception
  property show_usage : Bool = false
  def message # new method
    @message || @cause.try &.message
  end
end

a = RuntimeError.new("whoops")
pp! a
pp! a.message

b = CB::Error.new(cause: a)
pp! b
pp! b.message
```

abrightwell:
> You can create exceptions that wrap other ones with `cause:`. That might be handy here?

Perfect! Updated to this.

PR #38, opened 2022-05-26
Add stanza rendering to backup token output

fdr:
Although you can get the (one and only) relevant stanza name via
`pgbackrest info`, returning the stanza from `cb` is more convenient,
and allows for scoping the configuration overriding done by the
fragment produced by `--format=pgbackrest` more helpfully on average.

See https://pgbackrest.org/user-guide-rhel.html for details on
configuration precedence.


fdr:
Not as fundamental as the other changes, but I think this is slightly nicer.

abrightwell (approved):
MSTM. :+1:

PR #41, opened 2022-05-31
Refactor LogDestAdd.

abrightwell:
These changes refactor the LogDestAdd action and client call.  In
particular, the client was taking the action as an argument. This is the
only client method that did this. I believe it was an experiment on how
we might want to pass such params to the client. However, I think we've
somewhat settled on keeping these items decoupled in this respect. So,
we are just undoing that here.  Also, we take the opportunity to do some
minor refactoring around the naming of the other log destination
actions.


will (approved):

will:
Nice

PR #43, opened 2022-05-31
Refactor `login` to an `Action`.

abrightwell:
This refactor required a bit more than just moving the function from
`Program` to an action. In order to do this, we first need to
distinguish between actions that 'require' a client and those that
don't.

For example, `TeamList` is an action that requires access to the API and
therefore needs an client configured. However, `Login` simply collects
and persists the login credentials used to obtain a token. This action
does not require any API access.


src/cb/action.cr
```diff
@@ -2,21 +2,31 @@ require "log"
...
+
+  # APIAction performs some action utilizing the API.
+  abstract class APIAction < Action
+    Error = Program::Error
+
+    property client : Client
+
+    def initialize(@client, @input = STDIN, @output = STDOUT)
+      super(@input, @output)
+    end
```
abrightwell, line 29: So, I'm less certain of how crystal handles calling the initializer of a super class. I assume this is appropriate... however, I kind of feel like the `super` isn't necessary given the `@` in the param list. :shrug:  For the majority of cases, this isn't an issue, however, I found in some tests we were simply depending on the default for input/output and that would cause an issue.

src/cb/backup.cr
```diff
@@ -1,7 +1,7 @@
 require "./action"
 
 module CB
-  class BackupCapture < Action
+  class BackupCapture < APIAction
```
abrightwell, line 4: This is the vast majority of the changes below.

src/cb/action.cr
will:
```suggestion
    def initialize(@client, @input = STDIN, @output = STDOUT)
    end
```

The tests seem to pass for me with it like this? I think it'd be okay to omit the call to `super` due to the `@` in the param list like you mention

src/cb/action.cr
```diff
@@ -2,21 +2,31 @@ require "log"
 require "./client"
 
 module CB
+  # Action is the base class for all actions performed by `cb`.
   abstract class Action
-    Log   = ::Log.for("Action")
-    Error = Program::Error
```
will, line 7: I think we can leave the `Error =` up in the superclass?

src/cb/completion.cr
```diff
@@ -54,6 +54,8 @@ class CB::Completion
         firewall
       when "logdest"
         logdest
+      when "login"
+        suggest_none
```
will, line 58: In this section, we know there is a client so we don’t need them to log in. Maybe this should be the suggestion in the `rescue NoClientError` part?

src/cb/action.cr
abrightwell: Good catch, I thought I had moved it up also. Will fix.

src/cb/completion.cr
abrightwell: fixed.

will (approved):
Should we do a release after this?

PR #45, opened 2022-06-13
Token Action Refactor

abrightwell:
Two things here:

* Pull up the `Action` setter macros to the base `Action` class to make them available to all `Action`s.
* Move the `cb token` command from a `Proc` to an `Action`.

This does not change any functionality.  This is just another step in the direction of organizing the code a bit more.

### DIFF
--- a/src/cb/token.cr
+++ b/src/cb/token.cr
@@ -1,5 +1,31 @@
 require "./cacheable"
 
+# TODO (abrightwell): We had to explicitly qualify this class name as an
+# `Action` due to conflicts with the below `Token` struct.  Would be great to
+# potentially namespace actions under `CB::Action` or something. Something
+# perhaps worth considering.
+class CB::TokenAction < CB::Action
+  enum Format
+    Default
+    Header
+  end
+
+  property token : Token
+  property format : Format = Format::Default
+
+  def initialize(@token, @input, @output)
+  end
+
+  def run
+    case @format
+    when "header"
+      output << "Authorization: Bearer #{token.token}"
+    when "default"
+      output << token.token
+    end
+  end
+end
+
 struct CB::Token
   Cacheable.include key: host
   getter host : String
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -440,10 +440,9 @@ op = OptionParser.new do |parser|
 
   parser.on("token", "Return a bearer token for use in the api") do
     parser.banner = "cb token [-H]"
-    action = ->{ puts PROG.token.token }
-    parser.on("-H", "Authorization header format") do
-      action = ->{ puts "Authorization: Bearer #{PROG.token.token}" }
-    end
+    token = action = CB::TokenAction.new PROG.token, PROG.input, PROG.output
+
+    parser.on("-H", "Authorization header format") { token.format = CB::TokenAction::Format::Header }
   end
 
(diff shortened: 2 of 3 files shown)


src/cb/token.cr
```diff
@@ -1,5 +1,31 @@
 require "./cacheable"
 
+# TODO (abrightwell): We had to explicitly qualify this class name as an
+# `Action` due to conflicts with the below `Token` struct.  Would be great to
+# potentially namespace actions under `CB::Action` or something. Something
+# perhaps worth considering.
+class CB::TokenAction < CB::Action
```
abrightwell, line 7: I just wanted to call this one out. I'm not sure what the right solution would be here. However, I'm definitely feeling like we could start separating/organizing these things a little. But for now, as doing more would be out of scope of these changes, this seemed acceptable.

src/cb/token.cr
```diff
@@ -1,5 +1,31 @@
...
+  property token : Token
+  property format : String
+
+  def initialize(@token, @input, @output)
+    @format = "default"
+  end
+
+  def run
+    case @format
```
abrightwell, line 16: I can't imagine that there would be any other 'format'... as I can't initially think of any other form it might take but figured this was safe.  It also could play into some other todo list items I have about how we format output across the board, e.g. JSON vs table vs <insert here>... but that's a different discussion.

Could have definitely made this a `bool_setter` called `header` or something, which seems just as well and I would have no issue moving it to that.

src/cb/token.cr
```diff
@@ -1,5 +1,31 @@
...
+    when "header"
+      output << "Authorization: Bearer #{token.token}"
+    when "default"
+      output << token.token
+    else
+      # As it sits, it should not be possible to reach this point. However, for
+      # safety sake, it seemed worth the extra line. Just in case.
+      raise_arg_error("format", @format)
```
will, line 24: Instead of a string, format could be an enum maybe. It'll make the cli.cr part a bit more verbose, but then we wouldn't have to worry about unexpected values

https://crystal-lang.org/api/1.4.1/Enum.html

src/cli.cr
```diff
@@ -440,10 +440,9 @@ op = OptionParser.new do |parser|
 
   parser.on("token", "Return a bearer token for use in the api") do
     parser.banner = "cb token [-H]"
-    action = ->{ puts PROG.token.token }
-    parser.on("-H", "Authorization header format") do
-      action = ->{ puts "Authorization: Bearer #{PROG.token.token}" }
-    end
+    token = action = CB::TokenAction.new PROG.token, PROG.input, PROG.output
+
+    parser.on("-H", "Authorization header format") { token.format = CB::TokenAction::Format::Default }
```
will, line 445:
```suggestion
    parser.on("-H", "Authorization header format") { token.format = CB::TokenAction::Format::Header }
```

abrightwell: :man_facepalming:

src/cb/token.cr
will: Yeah, also the actions could maybe be a top level module instead of under CB, just to cut down on a bunch of typing all the time. It's mostly just for us, so whatever feels right

will (approved):

abrightwell:
Thanks!

PR #46, opened 2022-06-14
Move all actions to `CB::Action` module.

abrightwell:
This is not a functional change. Here we're only seeking to organize the
code a little bit further now that the action list has grown quite
extensively.

The impetus for this was originally stated as a comment in #45 about how we had a name conflict between `Token` and `TokenAction` and wanting to maybe clean that up a little.

Candidly, I'm not sure how I feel about this one overall. I do think that giving actions their own module is a good thing, however, I'm less confident in how I've accomplished that here.  More specifically, I toyed with the idea of `CB::Action` vs `Action`... and I tried both. I ultimately landed on `CB::Action` as it just felt a little better to me. However, I have no strong opinion here.

The other thing I did was I went through and move all the of the `CB::Action::<name>` declarations under a `module CB::Action` block. That resulted in a lot of 'files' appearing as changed given that it formatted them differently. However, the above is the only real change to these files.

I also opted to move the actions to their own directory. I don't think this is the final form as I wasn't certain what to do with the `scope_checks` directory. Seems like it might be appropriate to move it along with the `scope.cr` file.  Also, `dirs.cr` I was a little on the fence about, but felt it was best left where it was for now.

At any rate, this is just an initial swing at this but accepted that it could potentially use some polish and it might not be the only swing at it.


spec/cb/backup_spec.cr
```diff
@@ -42,26 +42,26 @@ private class BackupTestClient < CB::Client
 end
 
 private def make_ba
-  CB::BackupCapture.new(BackupTestClient.new(TEST_TOKEN))
+  CB::Action::BackupCapture.new(BackupTestClient.new(TEST_TOKEN))
```
will, line 45: One thing that might be worth trying, is if we `include CB::Action` once at the top of a spec, then I think everywhere in that file we can just have `BackupCapture`. It might cut down on a lot of the extra scoping and be okay enough for a given file at a time?

will (approved):

will:
I like grouping them in the directory for sure

Issue #47, opened 2022-06-14
Feature request: Authenticate with environment variable(s)


Taytay:
I would like to use the CLI as part of some automated tasks. We are currently writing to the file manually :
echo "$CRUNCHY_BRIDGE_AUTH" > $HOME/.config/cb/api.crunchybridge.com

I'd prefer to be able to set the credentials in a way similar to how heroku or aws CLIs let me do it though. For Heroku I can just set the `HEROKU_API_KEY` environment variables, and AWS has their similar `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.

abrightwell:
We are currently discussing some related improvements around setting and storing credentials. I'll add this to that discussion for consideration.

Taytay:
Thank you!

On a related note, I have just run across an instance where I appear to still be logged in (`cb whoami` returns a result) even though I've run `cb login` as someone else. I even tried removing the credentials stored in `~/.config/cb/api.crunchybridge.com`, but `cb whoami` still reports that I am logged in. This seems to indicate that the credentials are being picked up somewhere/somehow, but I don't know how I've managed this. Are the credentials ever stored anywhere outside of `~/.config/cb/api.crunchybridge.com`?

abrightwell:
The credentials are only stored in one place above. However, they are used to obtain a short lived access token. That token is stored in `~/.cache/cb/`. What's probably happening is that the token is still 'valid'. As a short term work around you could probably just remove the token from that directory and `cb` will automatically refresh it with the new credentials the next time you run it. I'll make a note to discuss adding a `logout` in the near term.

Taytay:
Ah, that's quite helpful. It was particularly confusing since I'd run `cb login` a second time and logged in with separate credentials, which I thought would do the trick. I appreciate the explanation and quick response.

will:
I know there are some limits on the api to the rate you can use the "big" creds to get the 1 hour cookie-like access token that you actually use for the api. 

I worry if we have env vars that use the big creds, people will quickly exhaust that limit. Maybe if we make `cb login` easier to use non-interactivly instead? Then a script would be like 

```
cb login <creds> 
cb dosomestuff
cb logout
```

Taytay:
Thank you for pointing that out! 

[...]

PR #49, opened 2022-06-15
Add `cb logout` command.

abrightwell:
These changes add a new command and action called `logout`. The purpose
of these are to allow a user to clear all credentials from their system
which has the effect of logging them out of the system.  To do this, we
remove both the stored credentials/API key. As well, we remove the
cached token regardless of whether or not it is expired.

In the case that none of the above items exist on the system, the effect
is the same as a no-op.

Mentioned in #47.

### DIFF
--- a/src/cb/creds.cr
+++ b/src/cb/creds.cr
@@ -34,4 +34,8 @@ struct CB::Creds
   def delete
     File.delete CONFIG/host
   end
+
+  def self.delete(host)
+    File.delete(CONFIG/host) if File.exists?(CONFIG/host)
+  end
 end
--- a/src/cb/token.cr
+++ b/src/cb/token.cr
@@ -41,6 +41,10 @@ struct CB::Token
     fetch? host
   end
 
+  def self.delete(host)
+    File.delete(file_path(host)) if File.exists?(file_path(host))
+  end
+
   def expires_at : Time
     Time.unix expires
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `cb logout` to logout a user from the CLI.
+
 
 ## [2.1.0] - 2022-06-03
 ### Added
--- a/src/cb/completion.cr
+++ b/src/cb/completion.cr
@@ -87,6 +87,7 @@ class CB::Completion
       "--help\tShow help and usage",
       "version\tShow version information",
       "login\tStore API key",
+      "logout\tRemove stored API key and token",
       "token\tGet current API token",
       "list\tList clusters",
       "team\tManage teams",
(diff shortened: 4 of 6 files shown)


src/cb/token.cr
```diff
@@ -41,6 +41,11 @@ struct CB::Token
     fetch? host
   end
 
+  def self.delete(host)
+    token = fetch host
+    token.delete unless token.nil?
```
abrightwell, line 46: I kind of feel like `Token#delete` and also `Creds#delete` should perhaps be more fault tolerant in the long run. However, I figured since it's not `nil` then it 'should' exist and should be safe. That doesn't mean I don't still wish it to be more fault tolerant, but I also don't want to explicitly silence any exceptions... not yet at least without good reason.

src/cb/creds.cr
```diff
@@ -34,4 +34,9 @@ struct CB::Creds
   def delete
     File.delete CONFIG/host
   end
+
+  def self.delete(host)
+    creds = for_host host
+    creds.delete unless creds.nil?
```
will, line 40: Just a few different ways to do this

```crystal
creds.delete unless creds.nil?
creds.delete if creds
creds.try &.delete
for_host(host).try &.delete # and drop the creds = line above
```

Any of them are fine, and it doesn't really matter, just wanted to point out the ways. I tend to like `unless` more when it is making a `nil?` go away rather than add it, like `if blah.nil?` to `unless blah`, but I'm sure i'm not consistent about that.

src/cb/cacheable.cr
```diff
@@ -22,15 +22,21 @@ module CB::Cacheable
     end
 
     def fetch?(key)
+      token = fetch key
```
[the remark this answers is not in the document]
will: I think in a earlier thing (not sure if ever committed) I had separate thing if it was expired vs missing. But then it just had the same logic, so it didn't matter the difference between the two cases. At that time at least.

The only thing with this change is that when there are pairs of methods in crystal, one with a `?` the convention I think is that the non question mark never returns a nilable object. This new non question mark one does. 

For the token delete, if the original `fetch?` would return nil, we would never use that token, so it's already as good as deleted in a sense. There may be a malformed file there, but cb wont use it.

So it might be okay to leave this method as it was? I'm not entirely sure. I think since the delete has a nil check already it's safe. The only difference would be an expired token will sit around on the file system, but that's ok?

src/cb/creds.cr
abrightwell: I think I like the last one most. I didn't like having to assign a var, but keep forgetting about that form.

src/cb/cacheable.cr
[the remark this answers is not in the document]
abrightwell: > So it might be okay to leave this method as it was? I'm not entirely sure. I think since the delete has a nil check already it's safe. The only difference would be an expired token will sit around on the file system, but that's ok?

So this is the impetus for my change here. If it's 'expired' then it doesn't matter, so effectively the same thing. But I was conflicted on whether or not it should be removed regardless. I'm not sure I have a strong opinion there, but it seemed reasonable to think all access related items would be removed. Another conflict I had was whether or not we just have a 'clear cache' kind of approach to it. :shrug: I don't want to over complicate it, but yeah... that's how I arrived at this implementation.

will (approved):

PR #52, opened 2022-06-23
try spectator

will:
I've been looking at it for a bit, but never actually tried to use it, and wanted to see what it'd look like. I did one file, the role thing. There are defiantly some nice things, especially coming from rspec which @fdr you are, but @abrightwell you aren't.

Take a look at the role spec before and after to see.

### DIFF
--- a/spec/cb/role_spec.cr
+++ b/spec/cb/role_spec.cr
@@ -1,15 +1,35 @@
 require "../spec_helper"
 include CB
 
-private TEST_ROLE = Client::Role.new name: "u_123", password: "secret", uri: URI.parse "postgres://u_123@foo.com"
+private TEST_ROLE = Client::Role.new(
+  account_email: "test@example.com",
+  name: "u_123",
+  password: "secret",
+  uri: URI.parse "postgres://u_123@foo.com",
+)
 
-CLUSTER = ClusterDetail.new(
+private SYSTEM_ROLE = Client::Role.new(
+  name: "application",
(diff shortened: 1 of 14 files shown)


abrightwell:
I think that I REALLY like this... when I first started contributing I remember thinking that having some kind of stubbing/mocking helpers would be huge. But I didn't have enough exposure, experience or knowledge at the time of what was available to or possible with crystal to take a swing at it. There is certainly a little bit of a learning curve for me, but one I'm very willing to climb if we decide to go this route. :+1:

spec/cb/role_spec.cr
```diff
@@ -1,121 +1,102 @@
...
+  major_version: 12,
+  plan_id: "memory-4",
+  cpu: 4,
+  memory: 111,
+  oldest_backup: nil,
+  provider_id: "aws",
+  region_id: "us-east-2",
+  network_id: "nfpvoqooxzdrriu6w3bhqo55c4",
+  storage: 1234
+)
```
abrightwell, line 42: I'm going to rethink how this is done. Seems like it would be preferable to have some kind of test data set. I played around with some ideas, but decided to defer it to a later PR.

brandur (approved):
TBH this is the first time I've even heard of Spectator so I'm probably not qualified to actually review this, but for what it's worth, seems okay to me. Selfishly, the DSL being the same as RSpec's makes things easier for me because I already know it.

PR #54, opened 2022-08-15
Add `cb role list` command.

abrightwell:
Add command to list a clusters roles.

`cb role list --cluster <cluster id>`

We also introduce a new dependency, of `epoch/tallboy` for more easily formatting output in table format. I had reviewed a few different libraries for this kind of thing and this one seemed to be the easiest to use as well as providing some more advanced features to include extending it for our purposes if necessary.  Initially I went with the `:ascii` renderer, but there I nothing preventing us from going a different direction with it if desired.

For roles `postgres` and `application` the associated account is set to `(system)`.  In the case of it being a user role, in other words starting with `u_` we display the associated account email.

Example:

```
> ./bin/cb role list --cluster ijpjdc2a4vhphhels3krggqhyu
+----------------------------------------------------------------+
| Cluster: Cluster 2022-05-05 00_26_59                           |
| Team:    personal                                              |
+------------------------------+---------------------------------+
| Role                         | Account                         |
+------------------------------+---------------------------------+
| application                  | (system)                        |
| postgres                     | (system)                        |
| u_sipm5b7unvehtadzukymk25ni4 | adam.brightwell@is-awesome.com  |
+------------------------------+---------------------------------+
```
```
> ./bin/cb role list --cluster ijpjdc2a4vhphhels3krggqhyu --format json
{
  "cluster": "Cluster 2022-05-05 00_26_59",
  "team": "personal",
  "roles": [
    {
      "role": "application",
      "account": "system"
    },
    {
      "role": "postgres",
      "account": "system"
    },
    {
      "role": "u_sipm5b7unvehtadzukymk25ni4",
      "account": "adam.brightwell@is-awesome.com"
    }
  ]
}


```

Note: This is a command that only works for users that are an admin of a team.

### DIFF
--- a/spec/cb/role_spec.cr
+++ b/spec/cb/role_spec.cr
@@ -4,28 +4,73 @@ private class RoleTestClient < CB::Client
   ACCOUNT = Account.new(
     id: "123",
     name: "user",
+    email: "test@example.com"
   )
 
-  ROLE = Role.new(
+  CLUSTER = ClusterDetail.new(
+    id: "pkdpq6yynjgjbps4otxd7il2u4",
+    team_id: "teamid",
+    name: "abc",
+    state: "na",
+    created_at: Time.utc(2016, 2, 15, 10, 20, 30),
+    is_ha: false,
+    major_version: 12,
+    plan_id: "memory-4",
+    cpu: 4,
+    memory: 111,
+    oldest_backup: nil,
+    provider_id: "aws",
+    region_id: "us-east-2",
+    network_id: "nfpvoqooxzdrriu6w3bhqo55c4",
+    storage: 1234
+  )
+
+  SYSTEM_ROLE = Role.new(
+    name: "application",
+    uri: URI.parse "postgres://application:secret@localhost:5432/postgres"
+  )
+
+  USER_ROLE = Role.new(
+    account_email: ACCOUNT.email,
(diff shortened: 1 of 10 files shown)


spec/cb/role_spec.cr
```diff
@@ -52,6 +92,42 @@ describe CB::RoleCreate do
...
+    action = CB::RoleCreate.new(RoleTestClient.new(TEST_TOKEN))
+
+    msg = /Missing required argument/
+
+    expect_cb_error(msg) { action.validate }
+    action.cluster_id = "pkdpq6yynjgjbps4otxd7il2u4"
+    action.validate.should eq true
+  end
+
+  it "#run" do
```
ngaumont, line 106: You might want to add another role in `ROLE` to check `system` is displayed.

ngaumont (changes requested):
This looks good to me, but I let others review the usage of tallboy.

spec/cb/role_spec.cr
abrightwell: Good idea. Done.

abrightwell:
Thanks @ngaumont. Along with the updates to the tests. I also added a `--format` flag so that output can be switched between a table and JSON.

will (approved):

abrightwell:
Thanks.

PR #55, opened 2022-08-22
Fix no output from `cb token`.

abrightwell:
Here we fix a bug with `cb token` that was preventing it from providing
any output.

The issue was that we were switching on a string for the format instead
of the enum for the format. This was due to a recent change on how the
format is set via `cli.cr`.


will (approved):

abrightwell:
Thanks.

PR #57, opened 2022-09-09
Improve role name validation error message.

abrightwell:
Previously, when an invalid role name was provided to `role update` or `role destroy` our error message was simply `invalid role` with nothing else very helpful provided. Here we're improving that message by also providing the list of currently supported valid role name values.

Example:

```
> ./bin/cb role destroy --cluster 66zlqgtfhrg4vlrhzt6ptbys2y --name test
error: invalid role: 'test'. Must be one of: application, default, postgres, user
```


gitstashpop (approved):
Thanks for the fix!

abrightwell:
Thanks!

PR #58, opened 2022-09-12
Add API_NAME_PATTERN.

abrightwell:
The API has some restrictions on what is allowed for user provided names of resources, for instance team or cluster names.  So, here we are adding a pattern for validating such names.


brandur (approved):
Seems okay by me!

abrightwell:
Thanks!

PR #59, opened 2022-09-12
Add `Identifier` class.

abrightwell:
Adds a class for encapsulating identifiers for resources like a team or a cluster.

This identifier can be the form of an eid or an api name. Validation is peformed using the respective regex patterns.

This will come in handy with some upcoming improvements I'm working on to allow for providing cluster names and teams as identifiers to commands.

### DIFF
--- a/src/cb/action.cr
+++ b/src/cb/action.cr
@@ -27,6 +27,19 @@ module CB
       end
     end
 
+    macro cluster_identifier_setter(property)
+      property {{property}} : Hash(Symbol, Identifier) = Hash(Symbol, Identifier).new
+
+      def {{property}}=(str : String)
+        raise Program::Error.new "invalid cluster identifier" if str.empty?
+
+        parts = str.split '/'
+
+        @{{property}}[:team] =  Identifier.new(parts.shift) unless parts.size == 1
+        @{{property}}[:cluster] = Identifier.new(parts.shift)
+      end
+    end
+
     # For simple identifiers such as region names, or plan names where we
     # expect only lowercase, numbers, and -
     macro ident_setter(property)
--- a/src/cb/identifier.cr
+++ b/src/cb/identifier.cr
@@ -1,0 +1,19 @@
+module CB
+  class Identifier
+    def initialize(@value : String = "")
+      raise Program::Error.new "invalid identifier: '#{@value}'" unless eid? || api_name?
+    end
+
+    def eid?
+      EID_PATTERN.matches? @value
+    end
+
+    def api_name?
+      API_NAME_PATTERN.matches? @value
+    end
+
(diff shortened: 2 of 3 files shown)


src/cb/action.cr
```diff
@@ -27,6 +27,19 @@ module CB
       end
     end
 
+    macro cluster_identifier_setter(property)
```
abrightwell, line 30: Allows for providing an identifier that can be namespaced as `<team id>/<cluster id>`.  But it also allows for just `<cluster id>`.

I'm scratching my head on making it 'simpler' to retrieve either component... but haven't settled on a solution just yet. But for now, this seems to be working quite well in my initial efforts.

will (approved):
I think this is okay. One thing though is what happens if someone names their team or cluster something that happens to match the EID pattern. It could either be an EID of nothing (in which case it'd probably just error?) or the name could happen to be the EID of another cluster or something (in which case maybe it'd silently use the wrong thing? is it even 'wrong' if it's that ambiguous?) 

It's probably not worth the effort to do anything about that situation, just that we think through what would happen.

Should platform and/or owl prevent user names from being eid patterns?

src/cb/identifier.cr
```diff
@@ -0,0 +1,19 @@
...
+      EID_PATTERN.matches? @value
+    end
+
+    def api_name?
+      API_NAME_PATTERN.matches? @value
+    end
+
+    def to_s
+      @value
+    end
```
will, line 17:
```suggestion
    def to_s(io : IO)
      io << @value
    end
```

https://crystal-lang.org/reference/1.5/guides/performance.html#dont-create-intermediate-strings-when-writing-to-an-io

abrightwell: Doh! I knew better... will fix. :+1:

abrightwell:
> I think this is okay. One thing though is what happens if someone names their team or cluster something that happens to match the EID pattern. It could either be an EID of nothing (in which case it'd probably just error?) or the name could happen to be the EID of another cluster or something (in which case maybe it'd silently use the wrong thing? is it even 'wrong' if it's that ambiguous?)

Yeah, definitely a consideration. I think if the name were match an eid but wasn't, that it would just 404/error. I don't think it would go silent. Or at least I would hope that it wouldn't.

> It's probably not worth the effort to do anything about that situation, just that we think through what would happen.

Originally, I didn't differentiate between the two. Simply assuming it was possible for both `eid?` and `api_name?` to be `true` for the same value. I'm honestly not sure what's right there. Perhaps worth a quick discussion.

> Should platform and/or owl prevent user-supplied names from being eid patterns?

That was my initial thought. But I'm not sure...

PR #60, opened 2022-09-13
Add spec `Factory`.

abrightwell:
Here we're simply adding a Factory for generating common objects that are the result of client requests, etc for testing purposes.

### DIFF
--- a/spec/factory.cr
+++ b/spec/factory.cr
@@ -1,0 +1,39 @@
+module Factory
+  extend self
+
+  def cluster(**params)
+    params = {
+      id:            "pkdpq6yynjgjbps4otxd7il2u4",
+      host:          "p.pkdpq6yynjgjbps4otxd7il2u4.example.com",
+      team_id:       "l2gnkxjv3beifk6abkraerv7de",
+      name:          "abc",
+      state:         "na",
+      created_at:    Time.utc(2016, 2, 15, 10, 20, 30),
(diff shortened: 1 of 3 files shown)


spec/factory.cr
```diff
@@ -0,0 +1,39 @@
...
+      major_version: 12,
+      plan_id:       "memory-4",
+      cpu:           4,
+      memory:        111,
+      oldest_backup: nil,
+      provider_id:   "aws",
+      region_id:     "us-east-2",
+      network_id:    "nfpvoqooxzdrriu6w3bhqo55c4",
+      storage:       1234,
+    }.merge(params)
```
abrightwell, line 22: I figured this approach was the easiest way for use to have some 'defaults' such that in the majority case we can just call `Factory.cluster`... however, if we needed multiple with different values for whatever reason, those could be overridden via kwarg, so `Factory.cluster id: abc123, name: foobar` but all well might be the same.

gitstashpop (approved):
lgtm. Would TestFactory be a more appropriate module name?

abrightwell:
> lgtm. Would TestFactory be a more appropriate module name?

Maybe? I'm not sure... I think given it's part of the `spec` package that it's implied. :man_shrugging: I'm not sure I have any strong opinions on it.

will (approved):

abrightwell:
Thanks!

PR #62, opened 2022-09-14
Allow `cb uri` to receive team and cluster names.

abrightwell:
Here we're adding the ability to provide a `name` for both a team and a cluster for the `cb uri` command.

The command continues to support providing an id for both as well.

### DIFF
--- a/src/cb/role.cr
+++ b/src/cb/role.cr
@@ -3,7 +3,6 @@ require "tallboy"
 
 abstract class CB::RoleAction < CB::APIAction
   eid_setter cluster_id
-  role_setter role
 end
 
 # Action to create a cluster role for the calling user.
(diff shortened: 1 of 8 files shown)


src/cb/role.cr
```diff
@@ -2,8 +2,18 @@ require "./action"
 require "tallboy"
 
 module CB
-  # Valid cluster role names.
   VALID_CLUSTER_ROLES = Set{"application", "default", "postgres", "user"}
```
abrightwell, line 5: I'm going to remove this in a follow up PR in favor of the new addition. I just didn't want to make this PR any more noisy than it has to be.

The idea is that I feel like 'roles' are complex enough to justify some of their own functions.  Whether that means they should be in their own module/domain or simply represented by a struct or class, I'm not sure. However, this felt like a good enough compromise at the moment. I'm not passionate about the solution so definitely open to better suggestions if one exists.

will (approved):
looks good

PR #65, opened 2022-09-22
Allow `cb psql` to receive team and cluster name.

abrightwell:
Here we're adding the ability to provide a `name` for both a team and a cluster for the `cb backup [capture | list | token]` commands.

The commands continues to support providing a cluster id.


abrightwell:
Thanks!

PR #66, opened 2022-09-23
Reorganize `CB::Client`.

abrightwell:
This is not a 'functional' change. It's just meant to help better organize the client code a bit as it has been growing considerably as of late. Here, we're breaking out all the 'resource' specific implementation into separate files. For instance, all `account` related response models and methods are defined in `account.cr`, the same for `cluster` implementation being found in `cluster.cr`.

We also move all of these files into their own directory at `src/client`.


abrightwell:
@will thoughts on this one?  I'm just trying to do some organization of the code.

will (approved):
Yeah nice I think splitting them out makes sense

abrightwell:
Thanks.

PR #67, opened 2022-09-24
More name-or-id updates.

abrightwell:
This is a WIP.  I'll be continuing to push to it until complete.  However, I wanted to get this up for early feedback.

~Based off of #66~

### DIFF
--- a/spec/cb/logs_spec.cr
+++ b/spec/cb/logs_spec.cr
@@ -1,0 +1,25 @@
+require "../spec_helper"
+
+Spectator.describe CB::Logs do
+  subject(action) { described_class.new client: client, output: IO::Memory.new }
+
+  let(client) { Client.new TEST_TOKEN }
+  let(cluster) { Factory.cluster }
+
+  describe "#validate" do
+    it "ensures required arguments are present" do
+      expect(&.validate).to raise_error Program::Error, /Missing required argument/
+
+      action.cluster_id = cluster.id
+      expect(&.validate).to be_true
+    end
+  end
+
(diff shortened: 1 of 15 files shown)


spec/cb/logs_spec.cr
```diff
@@ -0,0 +1,25 @@
...
+    end
+  end
+
+  describe "#call", skip: "Not implemented yet" do
+    # TODO (abrightwell): Testing #call for this action requires a considerable
+    # amount of mocking and adavanced setup given the call out to SSH2 libs and
+    # the nature of those. Also, before taking that on, it would be nice to update
+    # Spectator to the latest version. So, as to not make this changeset any
+    # noiser than necessary, we'll follow up on this one in a future PR.
+  end
```
abrightwell, line 24: I just wanted to call this out. I spent some time on Friday and Saturday trying to mock this one out in a way that intelligent. However, it was proving quite difficult. I then realized that we're on `0.10.5` of Spectator and `0.11.3` is the current version. I updated and found that while it eased some of the pains with this particular case, it has enough breaking changes in it that all other tests need to be updated accordingly. So, I figured I'd just handle that in a separate PR.

abrightwell:
I'm going to hold on making this PR any larger than it is for now. Once this is merged I'll follow up with similar updates to other commands.

spec/cb/logs_spec.cr
abrightwell: Also, FWIW, my goal in the near term is to get these kinds of things to a point where minimal line changes are necessary for such things in the future.

PR #68, opened 2022-09-26
Update `Spectator` to `0.11.3`.

abrightwell:
This is a pretty large changeset due in part to the breaking changes that were introduced by this update.

We also take the opportunity to go ahead and transition the remaining tests to bring them in line.

There were some interesting challenges. Mostly around how the create cluster action behaves. Specifically, it's the related client calls that caused a bit of a head ache. For some reason, Spectator doesn't like how we were doing it. And when stubbing the `create_cluster` method, things 'worked' (verified by tracing the call), but for some reason it would cause a crash elsewhere that was quite annoying. I believe it's possible that there is an issue in Spectator around this. However, I was unable to pin point it. However, I'm not terribly concerned with it as making the transition on these client methods has been on the TODO list for some time. So, we're also taking that opportunity with these changes.

### DIFF
--- a/spec/cb/cluster_create_spec.cr
+++ b/spec/cb/cluster_create_spec.cr
@@ -1,234 +1,172 @@
 require "../spec_helper"
 
-private class ClusterCreateTestClient < CB::Client
-  def create_cluster(arg)
-    Cluster.new("abc", "def", "my cluster", [] of Cluster)
-  end
-
-  def get_cluster(id : String)
-    ClusterDetail.new(
-      id: id,
-      team_id: "teamid",
-      name: "source cluster",
-      state: "na",
-      created_at: Time.utc(2016, 2, 15, 10, 20, 30),
-      host: "p.#{id}.test.crunchybridge.com",
-      is_ha: false,
-      major_version: 12,
-      plan_id: "memory-4",
-      cpu: 4,
-      memory: 111,
-      oldest_backup: nil,
-      provider_id: "aws",
-      region_id: "us-east-2",
-      maintenance_window_start: nil,
-      network_id: "nfpvoqooxzdrriu6w3bhqo55c4",
-      storage: 1234
-    )
-  end
-end
-
-private def make_cc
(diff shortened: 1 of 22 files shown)


spec/cb/cluster_create_spec.cr
```diff
@@ -1,233 +1,172 @@
...
+      action.plan = "plan"
+      expect_missing_arg_error
+      action.platform = "aws"
+      expect_missing_arg_error
+      action.region = "us-east-1"
+      expect_missing_arg_error
+      action.team = team.id
+
+      expect(&.validate).to be_true
+    end
```
abrightwell, line 25: I intend on coming back to this one. I was having some issues with passing the arg name to the macro. So, I figured I'd come back to it.

src/client/cluster.cr
```diff
@@ -69,19 +69,19 @@ module CB
...
+      #   is_ha:               cc.ha,
+      #   name:                cc.name,
+      #   plan_id:             cc.plan,
+      #   provider_id:         cc.platform,
+      #   postgres_version_id: cc.postgres_version,
+      #   region_id:           cc.region,
+      #   storage:             cc.storage,
+      #   team_id:             cc.team,
+      #   network_id:          cc.network,
+      # }
```
abrightwell, line 83: I'm thinking about extract this to a struct `CreateClusterParams` or something. Which could be used for all of `create`, `replica` and `fork`.

spec/cb/psql_spec.cr
```diff
@@ -4,18 +4,16 @@ include CB
 Spectator.describe CB::Psql do
   subject(action) { described_class.new client: client, output: IO::Memory.new }
 
-  let(client) { Client.new TEST_TOKEN }
+  mock_client do
+    def get_team_cert(team : String)
+      ""
+    end
```
abrightwell, line 10: Just overriding this method for now. As it's not important for the 'tests' and it just causes problems otherwise.

PR #69, opened 2022-10-13
Fix `ClusterDetail` cpu and memory type.

abrightwell:
The API recently updated from Int to Float. This should fix all related issues.

~Based on #70 until merged.~


will (approved):
This is good to fix the api change but I wonder if there is an easy way to not do .0 when it is zero

abrightwell:
> This is good to fix the api change but I wonder if there is an easy way to not do .0 when it is zero

I'll poke at it for a few minutes to see if I can make that work before merging. :+1:

PR #70, opened 2022-10-14
Upgrade to `ameba` 1.1.0.

abrightwell:
Crystal 1.6.0 is now available. But that causes issue with `ameba`. Unfortunately, 1.5.1 is no longer available from homebrew... or at least not easily. However, it seems like going from `ameba` 1.0.0 to 1.1.0 seems to resolve the issue, for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,16 +48,11 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    container: crystallang/crystal:latest-alpine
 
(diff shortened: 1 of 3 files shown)


abrightwell:
@will thoughts on this?

will (approved):
Yeah, I think going up to 1.6 is great. The only potential probelm might be the github automation for builds, but we can see how that shakes out when we try

abrightwell:
Thanks!

PR #75, opened 2022-11-03
Add maintenance on info

ngaumont:
They are quite a lot of information for this option, but I tried to give the more inputs to the end user.


spec/cb/cluster_info_spec.cr
```diff
@@ -35,17 +35,18 @@ Spectator.describe CB::ClusterInfo do
...
+              state: na
+               host: p.pkdpq6yynjgjbps4otxd7il2u4.example.com
+            created: 2016-02-15T10:20:30Z
+               plan: memory-4 (111GiB ram, 4vCPU)
+            version: 12
+            storage: 1234GiB
+                 ha: off
+           platform: aws
+             region: us-east-2
+        maintenance: no window set. Operations such as disk scaling, resize and postgres upgrade happens as soon as possible.
```
abrightwell, line 47: I'm curious if this should be `maintenance window` instead of just `maintenance`. I don't have a strong opinion here, my only thought is that perhaps it should be more explicit to avoid potential confusion.

src/cli.cr
```diff
@@ -237,6 +237,27 @@ op = OptionParser.new do |parser|
...
+
+    parser.on("info", "Detailed cluster maintenance information") do
+      info = set_action MaintenanceInfo
+      parser.banner = "cb maintenance info <--cluster>"
+      parser.on("--cluster ID", "Choose cluster") { |arg| info.cluster_id = arg }
+    end
+
+    parser.on("update", "Update cluster maintenance") do
+      update = set_action MaintenanceUpdate
+      parser.banner = "cb maintenance update <--cluster>"
```
abrightwell, line 255: If `--window-start` is required then let's list it in the banner.

```suggestion
      parser.banner = "cb maintenance update <--cluster> <--window-start>"
```

src/cli.cr
```diff
@@ -237,6 +237,27 @@ op = OptionParser.new do |parser|
...
+      parser.on("--cluster ID", "Choose cluster") { |arg| update.cluster_id = arg }
+      parser.on("--window-start START", "Hour maintenance window start (UTC). -1 to unset") { |arg| update.window_start = arg }
```
abrightwell, line 257: I know that the API uses `-1` to reset. But I'm curious if we could perhaps create a better experience for the user. For instance. Perhaps instead we either 1) offer another option called `--unset` or maybe we provide an enum of valid values for `--window-start` where something like `NONE` or similar would be translated to -1. :thinking:  It also might be a good idea to do some checks to ensure valid input for this option as well, None,0-23... or whatever.

abrightwell (changes requested):
A couple comments and suggestions.

Also, I'm curious if it would make sense to perhaps break out `MaintenanceWindow` into it's own class. That way the `action` is purely an action/command.

src/cb/maintenance.cr
```diff
@@ -0,0 +1,66 @@
+require "./action"
+
+abstract class CB::MaintenanceAction < CB::APIAction
+  cluster_identifier_setter cluster_id
+
+  def validate
+    check_required_args { |missing| missing << "cluster" if @cluster_id.empty? }
+  end
+
+  def self.window_start_explanation(maintenance_start : Int32?, now : Time? = nil)
```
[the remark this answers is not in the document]
ngaumont: This can't be an instance method, as it's also called in `cluster_info.cr`.
I did not find any better solution than class method …

src/cb/maintenance.cr
```diff
@@ -0,0 +1,66 @@
...
+    if maintenance_start.nil?
+      return "no window set. Operations such as disk scaling, resize and postgres upgrade happens as soon as possible."
```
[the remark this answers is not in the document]
ngaumont: Maybe no so verbose, but like this ?

```
no window set. Operations happens as soon as possible
```
I fear that the user might think that nothing will happen to his cluster because there is no window.
If I'm the only one with the fear, I'm good with only no window set.

spec/cb/cluster_info_spec.cr
ngaumont: I first used `maintenance window`, but this ended being very long with the description I had. With a shorter description, it fit.

src/cb/maintenance.cr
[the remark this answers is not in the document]
will: Maybe show `00:00-23:59` or something similar to indicate it's the entire day when there is no window set?

src/cb/maintenance.cr
```diff
@@ -0,0 +1,70 @@
...
+# Action to update cluster maintenance window
+class CB::MaintenanceUpdate < CB::MaintenanceAction
+  i32_setter window_start
+  bool_setter unset
+
+  def validate
+    super
+
+    raise Error.new "Must only use '--window-start' or '--unset' but not both." if window_start && unset
+    raise Error.new "window-start should be between 0 and 23" if (start = window_start) && (start < 0 || start > 23)
```
abrightwell, line 59:
```suggestion
    raise Error.new "Must use '--window-start' or '--unset' but not both." if window_start && unset
    raise Error.new "'--window-start' should be between 0 and 23" if (start = window_start) && (start < 0 || start > 23)
```

abrightwell (changes requested):
One minor adjustment. Otherwise, LGTM.

ngaumont:
> Also, I'm curious if it would make sense to perhaps break out `MaintenanceWindow` into it's own class. That way the `action` is purely an action/command.

I create a fixup and let you decide what you prefer

abrightwell (approved):

PR #77, opened 2022-11-08
Fix cb restart

ngaumont:


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add and remove a cluster to a Tailscale network with `cb tailscale (dis)connect`
 
+### Fixed
+- Fix cluster id bug with `cb restart`
+
```
abrightwell, line 13:
```suggestion
- Fix cluster id bug with `cb restart`
- Fix API route bug with `cb detach`
```

abrightwell (approved):
LGTM. One minor update suggestion.

CHANGELOG.md
abrightwell: I'll do a release once done.

ngaumont: Just to clarify, `cb detach` is working before this patch, and changing the route is only to use the real end point and not some redirection.

abrightwell: Ok, sounds good. Ignore the suggestion then. :-)

PR #78, opened 2022-11-15
Info and cancel on maintenance

ngaumont:
This PR updates `cb upgrade` to be more abstract so that it can be used inside maintenance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - `cb upgrade status` returns maintenance window information
+- `cb maintenance` command now supports `cancel` 
 
 ## [3.1.0] - 2022-11-18
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `cb upgrade status` returns maintenance window information
 
 ## [3.1.0] - 2022-11-18
 ### Added
--- a/spec/cb/cluster_upgrade_spec.cr
+++ b/spec/cb/cluster_upgrade_spec.cr
@@ -59,6 +59,43 @@ Spectator.describe CB::UpgradeStatus do
 
     expect(&.output.to_s).to eq expected
   end
+
+  it "#run display ha operation by default" do
+    action.cluster_id = cluster.id
+
+    expect(client).to receive(:get_cluster).and_return(cluster)
+    expect(client).to receive(:get_team).and_return(team)
+    expect(client).to receive(:upgrade_cluster_status).and_return([CB::Client::Operation.new("ha_change", "fake")])
+
+    action.call
+
+    expected = <<-EXPECTED
+    #{team.name}/#{cluster.name}
+      maintenance window: no window set. Default to: 00:00-23:59
+               ha_change: fake\n
+    EXPECTED
+
+    expect(&.output.to_s).to eq expected
+  end
+
+  it "#run filter ha operation if told so" do
+    action.cluster_id = cluster.id
+    action.maintenance_only = true
+
+    expect(client).to receive(:get_cluster).and_return(cluster)
+    expect(client).to receive(:get_team).and_return(team)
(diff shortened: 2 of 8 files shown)


spec/cb/cluster_upgrade_spec.cr
```diff
@@ -96,7 +96,7 @@ Spectator.describe CB::UpgradeStatus do
 
     action.call
 
-    output.to_s.should eq "personal/source cluster\n  no upgrades in progress\n"
+    output.to_s.should eq "personal/source cluster\n  no operations in progress\n  maintenance window: no window set. Default to: 00:00-23:59\n"
```
abrightwell, line 99:
```suggestion
    expected = <<-EXPECTED
    personal/source cluster
      no operations in progress
      maintenance window: no window set. Default to: 00:00-23:59\n
    EXPECTED
    
    expect(&.output.to_s).to eq expected
```

CHANGELOG.md
```diff
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Add and remove a cluster to a Tailscale network with `cb tailscale (dis)connect`
-- New command `cb maintenance info|update` to manage cluster maintenance window
-- `cb info` now also returns some maintenance window information
+- New command `cb maintenance info|update|cancel` to manage cluster maintenance window
+- `cb info`| `cb upgrade status` now also return some maintenance window information
```
abrightwell, line 11:
```suggestion
- `cb info` and `cb upgrade status` now return maintenance window information
```

src/cb/completion.cr
```diff
@@ -328,29 +328,20 @@ class CB::Completion
   def maintenance
     case @args[1]
     when "info"
-      maintenance_info
+      upgrade_status
```
abrightwell, line 331: I feel that I'd like to keep these separate for the time being.  The reason being that other options might be made available on maintenance info in the future. Though, I could see the argument for simply updating accordingly once that time comes if ever.

However, if not keeping separate, then I think it would be best to rename `upgrade_status` to something else that either 1) makes it more 'generic' or 2) indicates that it's also for `maintenance_info`.

abrightwell (commented):
Overall I'm not sure I understand this one. Are maintenance windows considered an 'upgrade'?  In my view, it doesn't seem that they are. And so, I think there is value in keeping these concepts separate from a certain perspective.

I do see the value in trying to DRY up or reuse some of the code, but I'm not sure this makes sense to me as it sits. I'd prefer that we keep these concepts separate and refactor what we can to make better reuse if possible.

Thoughts?

src/cb/completion.cr
abrightwell: I guess I'm still curious why `maintenance info` would render the same information as `upgrade status`. It just makes me curious why `maintenance info` is even necessary if they are the same.

abrightwell: Ok, this is making more sense to me now. :+1:

abrightwell (approved):

abrightwell:
Just a few changes requested, otherwise LGTM. :+1:

src/cb/cluster_upgrade.cr
```diff
@@ -51,28 +51,43 @@ class CB::UpgradeCancel < CB::Upgrade
...
-        output << op.flavor.rjust(pad).colorize.bold << ": "
-        output << op.state << "\n"
-      end
+      output << "  no #{operation_kind} in progress\n".colorize.bold
+    end
+
+    pad = (details.keys.map(&.size).max || 8) + 2
+    details.each do |k, v|
+      output << k.rjust(pad).colorize.bold << ": "
+      output << v << "\n"
```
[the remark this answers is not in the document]
ngaumont: Is there any difference between the two versions `'` and `"`?
Should `": "` also be changed to `': '` ?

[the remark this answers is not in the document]
abrightwell: So, this is something that came up in a previous view of mine. @will pointed out that it likely doesn't matter as llvm might potential optimize it out. However, for consistency and convention I've been trying to ensure that when using a single character to just make it a `Char` vs. `String`. Without digging too deep on it, I don't have a strong opinion on what's best, which is why I'm just falling back on 'convention'.  Admittedly, we're not 100% consistent across the code base, but as I encounter these I've been trying to update them.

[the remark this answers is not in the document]
abrightwell: Sorry, I didn't answer your questions... Haha! I'll do that now... :man_facepalming: 

> Is there any difference between the two versions `'` and `"`?

Yeah, one is `'` --> `Char` and `"` -> `String`.

> Should ": " also be changed to ': ' ?

Nope, wouldn't be possible. :-)

PR #80, opened 2022-12-01
Fix `--network` not being honored by `cb create --replica`.

abrightwell:
Previously, `--network` was an allowed option when creating a replica. At the time it wasn't supported by the API and shouldn't have been a valid option. However, since then the API now supports it. So, here we're simply fixing `cb create --replica` to honor the values it's passed for this option.  If no option is passed it will default to the source cluster.


abrightwell:
Thanks!

PR #83, opened 2022-12-09
Add `cb config-param` command.

abrightwell:
**WIP: Until API supports.**

Add `cb config-param` command.

Add `cb config-param` command to manage custom cluster configuration
parameters.

Supports the following subcommands:

`get` - get configuration parameter values.
`set` - set configuration parameter values.
`reset` - reset configuration parameter values.

Examples:
```
$ cb config-param get
Component   Name              Value
───────────────────────────────────────
postgres    max_connections   100
```

```
$ cb config-param set postgres:max_connections=200
 Component   Name              Value
───────────────────────────────────────
 postgres    max_connections   200
```

```
$ cb config-param reset --name postgres:max_connections
 Component   Name              Value
───────────────────────────────────────
 postgres    max_connections   100
```

Completes BR-973

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,7 @@ clean:
 .PHONY: spec
 spec: libs deps $(SOURCES)
 	$(CRYSTAL_BIN) tool format --check
+	$(CURDIR)/bin/ameba
 	@if [ "$(TARGET_OS)" == "darwin" ]; then \
 		LIBRARY_PATH=$(STATIC_LIBS_DIR) $(CRYSTAL_BIN) spec -Dmt_no_expectations --error-trace; \
 	else \
--- a/src/cb/config_param.cr
+++ b/src/cb/config_param.cr
@@ -1,0 +1,127 @@
+require "./action"
+
+module CB
+  # API action for configuration parameters.
+  #
+  # All configuration parameter actions must inherit this action.
+  abstract class ConfigurationParameterAction < APIAction
+    # The cluster ID.
+    cluster_identifier_setter cluster_id
+
+    # The output format. The default format is table format.
+    format_setter format
+
+    # Flag to indicate whether the output should include a header. This only
+    # has an effect when the output format is a table.
+    property with_header : Bool = true
+
+    # Result of api calls.
+    property parameters : Array(Client::ConfigurationParameter) = [] of Client::ConfigurationParameter
+
+    def validate
+      check_required_args do |missing|
+        missing << "cluster" if @cluster_id.empty?
+      end
+    end
+
+    def run
+      validate
+    end
+
+    protected def output_default
+      table = Table::TableBuilder.new(border: :none) do
+        columns do
+          add "Component"
+          add "Name"
+          add "Value"
+        end
+
+        header if with_header
+
+        @parameters.each do |p|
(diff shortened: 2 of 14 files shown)


src/cb/config_param.cr
```diff
@@ -0,0 +1,127 @@
...
+    name_setter? name
+
+    def run
+      super
+
+      if @name.nil?
+        @parameters = client.get_all_configuration_parameters cluster_id[:cluster]
+      else
+        @parameters = [client.get_configuration_parameter(cluster_id[:cluster], @name.as(String))]
+      end
```
ngaumont, line 68: I suppose you could switch the clause to remove nil.
Also, you can remove `@` as there is a property:
```suggestion
      if name
        parameters = [client.get_configuration_parameter(cluster_id[:cluster], name)]
      else
         parameters = client.get_all_configuration_parameters cluster_id[:cluster]
      end
```

src/cb/config_param.cr
```diff
@@ -0,0 +1,127 @@
...
+
+      case @format
```
ngaumont, line 70:
```suggestion
      case format
```

src/cb/config_param.cr
```diff
@@ -0,0 +1,127 @@
...
+  # Action for resetting configuration parameters.
+  class ConfigurationParameterReset < ConfigurationParameterAction
+    # The name of the parameter to reset.
+    name_setter? name
+
+    def run
+      super
+
+      if @name.nil?
+        @parameters = client.reset_all_configuration_parameters cluster_id[:cluster]
```
ngaumont, line 114: Do we want to allow that ?
Right now, there is only a reset on a component, not globally, but this could be added

Makefile
```diff
@@ -108,6 +108,7 @@ clean:
 .PHONY: spec
 spec: libs deps $(SOURCES)
 	$(CRYSTAL_BIN) tool format --check
+	$(CURDIR)/bin/ameba
```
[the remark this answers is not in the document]
abrightwell: Only in CI. I'm not sure this is the ideal solution here, but figured it was better to have it than not. The more important thing is the configuration update.

src/cb/config_param.cr
```diff
@@ -0,0 +1,127 @@
...
+      output << {
+        "parameters": @parameters,
+      }.to_pretty_json << '\n'
+    end
+  end
+
+  # Action for getting configuration parameters.
+  class ConfigurationParameterGet < ConfigurationParameterAction
+    # The name of the configuration parameter to get.
+    name_setter? name
```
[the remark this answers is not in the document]
abrightwell: Agreed. Full testing will account for this as well.

src/cb/config_param.cr
abrightwell: Yeah, not sure. I think having it for global and by component would be nice. But I'm not 100% certain on the use case just yet.

src/cb/config_param.cr
abrightwell: Good call on the clause. But `@` is necessary on `parameters`.

abrightwell:
Superseded by #133.

PR #84, opened 2022-12-15
Add examples section to option parser.

abrightwell:
Here we're just adding a simple example section to the option parser output.

An example for a command can be set using the `examples` property on the parser and might look something like the following:

```
parser.examples = <<-EXAMPLES
  $ cb some-command --cluster abc123 --some-other-flag "some value"
EXAMPLES
```

### DIFF
--- a/src/ext/option_parser.cr
+++ b/src/ext/option_parser.cr
@@ -1,6 +1,8 @@
 require "option_parser"
 
 class OptionParser
+  property examples : String?
+
   # Allows for hiding an option from help.
   def on(flag : String, &block : String ->)
     flag, value_type = parse_flag_definition(flag)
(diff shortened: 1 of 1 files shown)


ngaumont (approved):

abrightwell:
Thanks!

src/ext/option_parser.cr
```diff
@@ -26,14 +28,20 @@ class OptionParser
...
       io << '\n'
     end
 
     if !flags.empty?
       io << '\n' << "Options".colorize.bold << ":\n"
       flags.sort_by(&.lstrip(" -")).join io, '\n'
+      io << '\n'
+    end
+
+    unless examples.nil?
```
will, line 42: Not worth changing after this has been merged, but I'd prefer `if examples` in this situation

abrightwell: :man_facepalming: I know better. Haha! I'll add it to one of my current wips. :+1:

PR #85, opened 2022-12-23
Add `Format` enum and custom table renderer.

abrightwell:
These are the same changes that are introduced in #83.  However, I'm separating them out so that they can start to be incorporated in other actions/commands since that PR is still a WIP/Draft.

* Add `Format` enum for output formats.

Previously, we had started to define this at the action level as it was
more about experimenting with it. But now it seems like something we
should likely just pull up to the top level so that all actions can
benefit from it at some point.

* Add custom table renderer for Tallboy.

### DIFF
--- a/src/cb/format.cr
+++ b/src/cb/format.cr
@@ -1,0 +1,7 @@
+module CB
+  enum Format
+    Default
+    JSON
+    Table
+  end
+end
--- a/src/cb/table.cr
+++ b/src/cb/table.cr
@@ -1,0 +1,40 @@
+require "tallboy"
+
+module CB::Table
+  class TableBuilder < Tallboy::TableBuilder
+    def header
+      header(@columns.map(&.name.colorize.bold.underline))
+    end
+
+    def header(arr : Array)
+      row(arr, :none)
+    end
+
+    def render(io = IO::Memory.new)
+      DefaultRenderer.new(self.build).render(io)
+    end
+  end
+
+  class DefaultRenderer < Tallboy::Renderer
+    @@border_style = {
+      "corner_top_left"     => "",
+      "corner_top_right"    => "",
+      "corner_bottom_right" => "",
+      "corner_bottom_left"  => "",
+      "edge_top"            => "",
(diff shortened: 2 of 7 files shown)


src/cli.cr
```diff
@@ -475,7 +475,7 @@ op = OptionParser.new do |parser|
     parser.banner = "cb token [-H]"
     token = action = CB::TokenAction.new PROG.token, PROG.input, PROG.output
 
-    parser.on("-H", "Authorization header format") { token.format = CB::TokenAction::Format::Header }
+    parser.on("-H", "Authorization header format") { token.with_header = true }
```
abrightwell, line 478: I like this pattern here better.  We could obviously include it as part of the enum, however, I feel like this is a special case and should be exempt.

src/cb/format.cr
```diff
@@ -0,0 +1,7 @@
+module CB
+  enum Format
+    Default
```
abrightwell, line 3: I'm keeping `Default` around for the time being. It is likely going away as I'd like for the format setter to receive a default. But I'm unsure if it should be this implicitly or set explicitly by each action.

For instance, this works:

```
format_setter format

case @format
when Format::Default, Format::Table
when ...
end
```

But I'd prefer it to be something like:

```
format_setter format = Format::Table

case @format
when Format::Table
when ...
end
```

But I'll probably want to try to bring that in with a later change since this is meant to only refactor the existing from per action format enums to a module level format enum.

src/cb/table.cr
```diff
@@ -0,0 +1,40 @@
...
+
+module CB::Table
+  class TableBuilder < Tallboy::TableBuilder
+    def header
+      header(@columns.map(&.name.colorize.bold.underline))
+    end
+
+    def header(arr : Array)
+      row(arr, :none)
+    end
```
abrightwell, line 11: By setting `border` to `:none` we eliminate the extra line use as a separator.  And since we add an underline to the column header name (above) we keep the clear delineation between header and data.

PR #88, opened 2022-12-28
Add `cb network` command for managing networks.

abrightwell:
Supports the following commands:

```
> ./bin/cb network --help
Usage:
    cb network <info|list>

Commands:
    info                             Detailed network information
    list                             List all networks

Options:
```

```
> ./bin/cb network info --help
Usage:
    cb network info <--network>

Options:
    --format FORMAT                  Choose output format (default:
table)
    --network ID                     Choose network
    --no-header                      Do not display table header

Examples:
Get network details. Ouptut: table
$ cb network info --network <ID>

Get network details. Output: table without header
$ cb network info --network <ID> --no-header

Get network details. Output: json
$ cb network info --network <ID> --format=json
```

```
> ./bin/cb network list --help
Usage:
    cb network list [--team]

Options:
    --format FORMAT                  Choose output format (default:
table)
    --no-header                      Do not display table header
    --team ID                        Choose team

Examples:
List all networks for all teams with membership. Output: table
$ cb network list

List all networks for a specific team. Output: table
$ cb network list --team <ID>

List all networks for a specific team. Ouptut: table without header
$ cb network list --team <ID> --no-header

List all networks for a specific team. Output: json
$ cb network list --team <ID> --format=json
```

Examples:

```
$ cb network list
  ID                           Team                         Name                  CIDR4            Provider   Region
  oap3kavluvgm7cwtzgaaixzfoi   l2gnkxjv3beifk6abkraerv7de   Default Network       192.168.0.0/24   aws        us-east-1

$ cb network list --no-header
  oap3kavluvgm7cwtzgaaixzfoi   l2gnkxjv3beifk6abkraerv7de Default Network   192.168.0.0/24   aws   us-east-1

$ cb network list --format=json
{
  "networks": [
    {
      "cidr4": "192.168.0.0/24",
      "id": "oap3kavluvgm7cwtzgaaixzfoi",
      "name": "Default Network",
      "provider_id": "aws",
      "region_id": "us-east-1",
      "team_id": "l2gnkxjv3beifk6abkraerv7de"
    }
  ]
}
```

### DIFF
--- a/src/cb/completion.cr
+++ b/src/cb/completion.cr
@@ -56,6 +56,8 @@ class CB::Completion
         logdest
       when "maintenance"
         maintenance
+      when "network"
+        network
       when "psql"
         psql
       when "restart"
@@ -99,6 +101,7 @@ class CB::Completion
       "teamcert\tGet team public cert",
       "tailscale\tManage tailscale",
       "maintenance\tManage cluster maintenance",
+      "network\tManage networks",
       "info\tDetailed cluster info",
       "uri\tConnection uri",
       "create\tProvision a new cluster",
@@ -125,6 +128,11 @@ class CB::Completion
   def single_cluster_suggestion
     return cluster_suggestions if args.size == 2
     suggest_none
+  end
+
+  def team_suggestions
+    teams = client.get_teams
+    teams.map { |t| "#{t.id}\t#{t.name}" }
   end
 
   def cluster_suggestions
@@ -357,6 +365,50 @@ class CB::Completion
     suggest << "--cluster\tcluster id" unless has_full_flag? :cluster
     suggest << "--window-start\tmaintenance window start (UTC)" unless has_full_flag?(:unset) || has_full_flag?(:window_start)
     suggest << "--unset\tUnset mainetnance window" unless has_full_flag?(:unset) || has_full_flag?(:window_start)
+    suggest
+  end
+
+  def network
+    case @args[1]
+    when "info"
+      network_info
+    when "list"
+      network_list
+    else
+      [
+        "info\ndetailed network information",
+        "list\tlist available networks",
+      ]
+    end
+  end
+
+  def network_info
+    if last_arg?("--format")
+      return ["table", "json"]
+    end
+
+    if last_arg?("--network")
+      return [] of String
+    end
+
+    suggest = [] of String
+    suggest << "--format\tchoose output format" unless has_full_flag? :format
+    suggest << "--network\tnetwork id" unless has_full_flag? :network
+    suggest
+  end
+
+  def network_list
+    if last_arg?("--format")
+      return ["table", "json"]
+    end
+
+    if last_arg?("--team")
+      return team_suggestions
+    end
+
+    suggest = [] of String
+    suggest << "--format\tchoose output format" unless has_full_flag? :format
+    suggest << "--team\tchoose team" unless has_full_flag? :team
     suggest
   end
 
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -269,6 +269,59 @@ op = OptionParser.new do |parser|
   end
 
   #
+  # Network Management
+  #
+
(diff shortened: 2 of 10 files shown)


src/cb/completion.cr
```diff
@@ -360,6 +368,50 @@ class CB::Completion
...
+    if last_arg?("--network")
+      return [] of String
+    end
+
+    if last_arg?("--format")
+      return ["table", "json"]
+    end
+
+    suggest = [] of String
+    suggest << "--network\tntwork id" unless has_full_flag? :network
```
ngaumont, line 395:
```suggestion
    suggest << "--network\tnetwork id" unless has_full_flag? :network
```

src/cli.cr
```diff
@@ -268,6 +268,59 @@ op = OptionParser.new do |parser|
...
+      list = set_action NetworkList
+
+      parser.banner = "cb network list [--team]"
+
+      parser.on("--team ID", "Choose team") { |arg| list.team_id = arg }
+      parser.on("--format FORMAT", "Choose output format (default: table)") { |arg| list.format = arg }
+      parser.on("--no-header", "Do not display table header") { list.no_header = true }
+
+      parser.examples = <<-EXAMPLES
+      List all networks for all teams with membership. Output: table
```
ngaumont, line 309: If you are not a member of a team, you should not be to see anything.
I suppose the membership is implicitly implied.
```suggestion
      List all networks. Output: table
```

src/client/network.cr
```diff
@@ -0,0 +1,46 @@
...
+      return get_network id.to_s if id.eid?
+      get_network_by_name(id)
+    end
+
+    def get_network(id : String?)
+      resp = get "networks/#{id}"
+      Network.from_json resp.body
+    end
+
+    # Get all networks for a team.
```
ngaumont, line 26:
```suggestion
    # Get all networks for a team or all teams.
```

The documentation from https://crunchybridgeapi.docs.apiary.io/#reference/0/networks/list-networks also has to be updated

src/cb/completion.cr
```diff
@@ -360,6 +368,50 @@ class CB::Completion
...
+      [
+        "info\ndetailed network information",
+        "list\tlist available networks",
+      ]
+    end
+  end
+
+  def network_info
+    if last_arg?("--network")
+      return [] of String
```
ngaumont, line 387: Could you add a `network_suggestions` like for cluster or team ?

ngaumont (changes requested):
Sould I update `owl` to also return cidr6?

abrightwell:
> Sould I update `owl` to also return cidr6?

Not at this time I don't think. I think that there is a larger API discussion to be had around that one.

src/cb/completion.cr
abrightwell: I had considered that. However, since `list-networks` currently only returns for a single team (defaulting to the 'personal' team, I think), it's not possible to get ALL of the possible networks without creating a potentially large number of requests... 1 per team that an account is a member of... I really didn't like that approach. So I figured I'd just make this an 'improvement' in the future once we can improve it from the API perspective. I'll add a `TODO` here to ensure we don't miss it in the future.

Of course we **could** add a `--team` flag to this command and if present use it... or default to the personal for completions... but that just seems confusing to me. And part of me would prefer to improve this feature all the way down the chain instead.

src/cli.cr
abrightwell: Yeah, currently, the personal team is `implied`.  Perhaps not ideal. We COULD make `--team` a required flag. Which might clear up some potential confusion maybe.  What do you think?

abrightwell: Speaking with @jsoyland, we're going to keep `--team` optional. But I've updated the examples to reflect what to expect when it is omitted.

ngaumont (approved):

abrightwell:
Thanks.

PR #89, opened 2023-01-04
Add `Format::List`.

abrightwell:
Add an enum value for a list output format.

This format should typically be a two column 'key-value' style list.

For example:
```
  ID:   123abc
  Name: example
```
Also, as an example of it's usage, we update `TeamInfo` to use this enum. As well as update the action to use the `TableBuilder` to generate this output format.

### DIFF
--- a/src/cb/team.cr
+++ b/src/cb/team.cr
@@ -3,14 +3,18 @@ require "./action"
 abstract class CB::TeamAction < CB::APIAction
   eid_setter team_id
 
-  private def team_details(t : CB::Client::Team) : String
-    String.build do |str|
-      str << "ID:           \t" << t.id.colorize.t_id << "\n"
-      str << "Name:         \t" << t.name.colorize.t_name << "\n"
-      str << "Role:         \t" << t.role.to_s.titleize << "\n"
-      str << "Billing Email:\t" << t.billing_email << "\n"
-      str << "Enforce SSO:  \t" << (t.enforce_sso.nil? ? "disabled" : t.enforce_sso)
+  format_setter format
+
+  private def output_team_details(t : CB::Client::Team)
+    table = Table::TableBuilder.new(border: :none) do
+      row ["ID:", t.id.colorize.t_id]
+      row ["Name:", t.name.colorize.t_name]
+      row ["Role:", t.role.to_s.titleize]
+      row ["Billing Email:", t.billing_email]
+      row ["Enforce SSO:", (t.enforce_sso.nil? ? "disabled" : t.enforce_sso)]
     end
+
+    output << table.render << '\n'
   end
 end
 
@@ -43,7 +47,11 @@ end
 class CB::TeamInfo < CB::TeamAction
   def run
     team = client.get_team team_id
-    output << team_details(team) << "\n"
+
+    case @format
+    when Format::Default, Format::List
+      output_team_details(team)
+    end
   end
 end
 
@@ -80,7 +88,7 @@ class CB::TeamUpdate < CB::TeamAction
       "name"          => name,
     }
 
-    output << team_details(team) << "\n"
+    output_team_details(team)
   end
 end
 
--- a/spec/cb/team_spec.cr
+++ b/spec/cb/team_spec.cr
@@ -1,0 +1,29 @@
+require "../spec_helper"
+
+Spectator.describe TeamInfo do
+  subject(action) { described_class.new client: client, output: IO::Memory.new }
+
+  mock_client
+
+  let(team) { Factory.team }
+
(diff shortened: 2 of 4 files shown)


abrightwell:
@ngaumont thoughts on this one?

will:
@abrightwell I didn't notice this on the original table PR, but just as an aside, you might want to change `row` to be variadic instead of taking an array. The drawback is that we need to know at compile time how many columns there'd be in each row. If there are some spots where that is impossible, we could maybe have an overloaded implementation of `row` that still takes arrays like it does today. 

The difference is pretty big in benchmarking. It honestly probably doesn't noticeably matter in real life, for a program like this, but every little bit helps.

```crystal
require "benchmark"

def row(array)
  array
end

def row2(*args)
  args
end

p row ["one", "two"]
p row2 "one", "two"

Benchmark.ips do |x|
  x.report("array")  { row ["one", "two"] }
  x.report("vararg") { row2 "one", "two" }
end
``` 

```
crystal run --release a.cr
["one", "two"]
{"one", "two"}
 array  23.19M ( 43.11ns) (± 3.52%)  64.0B/op  22.98× slower
vararg 533.01M (  1.88ns) (± 5.92%)   0.0B/op        fastest
```

The array one needs to allocate memory on the heap for each array, in this case 64B. Which isn't a lot, but the `Tuples` are `Struct`s instead of `Class`es, so they are on the stack and don’t need to allocate any memory.

abrightwell:
Wow, that's interesting.  I'll follow up with some kind of improvement in a separate PR. But good to know. :+1:

abrightwell:
> Wow, that's interesting. I'll follow up with some kind of improvement in a separate PR and then update this one. But good to know. +1

I took a quick swing at this and it seems like to realize the benefits will require a more extensive update in overriding/customizing tallboy. I could be wrong, as it was just a quick attempt. But, I'm going to keep it on my list of things to dig into as time allows, as I agree every little bit counts.

src/cb/team.cr
```diff
@@ -3,14 +3,18 @@ require "./action"
...
-  private def team_details(t : CB::Client::Team) : String
-    String.build do |str|
-      str << "ID:           \t" << t.id.colorize.t_id << "\n"
-      str << "Name:         \t" << t.name.colorize.t_name << "\n"
-      str << "Role:         \t" << t.role.to_s.titleize << "\n"
-      str << "Billing Email:\t" << t.billing_email << "\n"
-      str << "Enforce SSO:  \t" << (t.enforce_sso.nil? ? "disabled" : t.enforce_sso)
+  format_setter format
+
+  private def output_team_details(t : CB::Client::Team)
```
ngaumont, line 8:
```suggestion
  private def output_team_details(t : CB::Client::Team)
    case @format
    when Format::Default, Format::List
      output_list_team_details(team)
    end
  end
  
  private def output_list_team_details(t : CB::Client::Team)
```

src/cb/team.cr
```diff
@@ -43,7 +47,11 @@ end
 class CB::TeamInfo < CB::TeamAction
   def run
     team = client.get_team team_id
-    output << team_details(team) << "\n"
+
+    case @format
+    when Format::Default, Format::List
+      output_team_details(team)
+    end
```
ngaumont, line 54:
```suggestion
    output_team_details(team)
```

ngaumont (changes requested):
Could you also update the commit message:

> As well, we introduce the use of the using the `TableBuilder` to create
> a list output. The output is the same. But the method is different in
> that by using the table builder we don't have to worry about layout of
> the values.
> 

As well, we introduce `TableBuilder` to create a list output.
The output is the same, but by using the table builder we don't have to worry
about the layout.

src/cb/team.cr
abrightwell: So, this change is intentional. While it doesn't necessarily matter at the moment, the intent is to demonstrate the intended usage of `Format::List`.  As well, it is setting the team commands up for future improvements around output formats.

src/cb/team.cr
abrightwell: Ultimately, we'll be refactoring `output_team_details` to something similar to what I've been doing on other commands:

For example what we've done in `role.cr`: https://github.com/CrunchyData/bridge-cli/blob/16591b1c363490803ee49b6701ed6f42934a91b8/src/cb/role.cr#L56-L61

But the intent here is to minimize the changes as they sit while still demonstrating the use of `Format::List`. Future improvements to the team command (and others) will be coming in future PR's.

ngaumont (approved):

abrightwell:
Thanks!

PR #90, opened 2023-01-06
Allow API key secret-only login and usage

brandur:
API keys have both an application ID and application secret, and
previously both of them were required to use one. At some point we
realized this hadn't really been entirely thought through and wasn't
really necessary since API key secrets already have plenty of entropy by
themselves, and as of mid-June last year we started to allow secret-only
authentication in the API. Newer secrets were prefixed with `cbkey_` so
that we could easily distinguish an API key secret from other types of
secrets that might be used for authentication (like an access token
secret).

Here, support secret-only authentication in cb. This makes `cb login` a
bit faster to use since you only need to copy in one value.

The downside to the approach I've built here is that we'd have to force
anyone with an older API key to rotate it to continue using a new
version of cb. I figure this might be okay though since we've been
issuing newer style API keys for seven months now, and a lot of people
will already have rotated theirs organically.

The other alternative is that we could continue to try and read an
application ID from `creds` but just stop storing one. When reading one,
allow ID to be empty on newer API keys, but use ID if an old one is
detected, and only error in the case there's no ID and the API key is
old. This would have improved backward compatibility, but make the code
a bit cruftier and more complex. I opted for the simpler alternative,
but don't mind implementing it this way if it's preferable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `cb upgrade status` returns maintenance window information
 - `cb upgrade start` command accepts `--starting-from` and `--now` options that
    specify upgrade failover window.
+- Specifying an application ID when adding an API key is no longer necessary. A
+  "prefixed" API key starting with `cbkey_` is necessary for use with cb. (All
+  new API keys are prefixed.)
 
 ### Fixed
 - Fix `--network` not being honored when passed with `cb create --replica`.
--- a/src/cb/creds.cr
+++ b/src/cb/creds.cr
@@ -5,12 +5,11 @@ struct CB::Creds
   include JSON::Serializable
 
   getter host : String
-  getter id : String
   getter secret : String
 
   CONFIG = Dirs::CONFIG
 
-  def initialize(@host, @id, @secret)
+  def initialize(@host, @secret)
   end
 
   def self.for_host(host) : Creds?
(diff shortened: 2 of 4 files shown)


brandur:
@will @abrightwell Thoughts on this one? Noticed while working on something else, but feels like we could streamline the process a bit.

abrightwell (commented):
I think this is fine by me.

Easing the login process is something that's certainly needed.

Though, I would be worried about breaking some users that have the process scripted.  So it might be worth continuing to support both in the way you mention. And adding a `deprecated` or `warn` message that the user should rotate their key.

Granted, there are enough significant changes on the horizon that the next release will most likely be a major one (IMO). And would make me less concerned about such a change. Therefore, if we don't continue to support app_id (which I think is a huge positive), then let's make sure to update the `README` as well, let's add an entry into the `CHANGELOG`.

brandur:
SG. Before adding compatibility, let's just try and objectively measure the number of old API keys in circulation with cb. Added https://github.com/CrunchyData/priv-all-platform/pull/2277 so that we can look for these easily.

brandur:
@abrightwell Okay, so added a probe earlier this week to see if people are using non-prefixed API keys, ... and I think we're okay. You can verify yourself by looking for `source:platform app:app[web] api_key_prefixed:false`.

So I'm kind of thinking maybe just not bother with the backwards compatibility here — thoughts? We could wait until Tuesday to be really sure, but I don't think anyone would end up taking advantage of it anyway.

abrightwell (approved):
SGTM. Let's add a mention of it to the CHANGELOG. 👍

brandur:
@abrightwell k, SG. I added a CHANGELOG entry. Going to merge this so we don't forget about it again.

PR #91, opened 2023-01-11
Implement command for `cb open` to open a Dashboard session

brandur:
Here, we implement a new command: `cb open`. It uses the API key bundled
into cb to create a session via API, then pops a web browser using a
system command like `open` on Mac or `xdg-open` on Linux to log the user
into the session via use of a one-time token (see [1] for background).

I'm not fully decided on the longer term survival of this feature (at
least with its current implementation) because although secure, it does
have does have to make a concession that's suboptimal for security.
Because commands like `open` can only support a `GET` request, one-time
tokens are sent via query parameter instead of body, which means they
could conceivably land in a log somewhere, even if the request itself is
operating over a secure channel.

We do take steps to keep one-time tokens secure -- they're named such
because they're always burned the first time the API ever sees one,
regardless of whether the request succeeds or not. So by the time it
ever leaks into a log, it should already be useless. One-time tokens are
also only usable within one minute of creation, so it's not possible to
not use on and have it linger around as a possible backdoor.

Still, it'd probably be overall better if more of this functionality was
pushed into the browser, but to make that happen, the Frontend would
have to be fully aware of login functionality, which is a much larger
project given its pace of change. So the idea here is to get something
in that works, and then come up with a more optimal fully-formed solution
later on.

In order to not make people too dependent on `cb open` in case we want
to change something, it's hidden from the `cb help` list so that it'll
generally only be used by those who already know about it.

[1] https://github.com/CrunchyData/priv-all-cloud-owl/pull/2252

### DIFF
--- a/spec/cb/open_spec.cr
+++ b/spec/cb/open_spec.cr
@@ -1,0 +1,46 @@
+require "../spec_helper"
+include CB
+
+Spectator.describe CB::Open do
+  subject(action) { described_class.new client: client, output: IO::Memory.new }
+
+  let(client_host) { "api.crunchybridge.com" }
+  let(session_id) { "af73g4rdcuyo3ol766fkmtfb3y" }
+  let(session_one_time_token) { "cbott_one_time_token_secret" }
+
+  mock_client
+
+  describe ".open_command" do
+    it "successfully returns the name of an executable" do
+      expect(action.class.open_command).to_not be_nil
+    end
+  end
+
+  describe "#call" do
+    before_each do
+      client.host = client_host
+    end
+
+    it "creates a session and executes open" do
+      open_args : Array(String)? = nil
+
+      action.open = ->(args : Array(String), _env : Process::Env) do
+        open_args = args
+        nil
+      end
+
+      expect(client).to receive(:create_session)
+        .with(
+          Client::SessionCreateParams.new(generate_one_time_token: true)
+        )
+        .and_return(
+          Client::Session.new(id: session_id, one_time_token: session_one_time_token)
+        )
+
+      action.call
+
+      expected_login_url = "https://#{client_host}/sessions/#{session_id}/actions/login?one_time_token=#{session_one_time_token}"
+      expect(open_args).to eq([expected_login_url])
+    end
+  end
+end
--- a/src/cb/open.cr
+++ b/src/cb/open.cr
@@ -1,0 +1,45 @@
+require "./action"
+
+module CB
+  # Uses the API key bundled into cb to open a Dashboard session for them by
+  # creating a session and opening a browser via appropriate system command.
+  class Open < APIAction
+    # Make the open command stub-able for testing purposes.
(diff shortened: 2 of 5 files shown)


brandur:
@abrightwell @will Hey, do you guys have any idea why the lint on this PR is failing? Basically I have a proc like this:

``` cb
      action.open = ->(args : Array(String), _env : Process::Env) do
        @open_args = args
        nil
      end
```

And it's telling me that the `args` argument isn't used:

```
Run bin/ameba
Inspecting 10[4](https://github.com/CrunchyData/bridge-cli/actions/runs/3888540565/jobs/6635966303#step:6:5) files

...............................................................................F........................

spec/cb/open_spec.cr:23:24
[W] Lint/UnusedArgument: Unused argument `args`. If it's necessary, use `_args` as an argument name to indicate that it won't be used.
> action.open = ->(args : Array(String), _env : Process::Env) do
```

This isn't true though. I've confirmed that the `args` variable is doing what it's supposed to be doing, and if you rename it to `_args`, the build unsurprisingly fails.

So I think this is a linter bug? What do you guys generally do to work around this kind of stuff?

abrightwell:
> @abrightwell @will Hey, do you guys have any idea why the lint on this PR is failing? Basically I have a proc like this:
> 
> ```
>       action.open = ->(args : Array(String), _env : Process::Env) do
>         @open_args = args
>         nil
>       end
> ```
> 
> And it's telling me that the `args` argument isn't used:
> 
> ```
> Run bin/ameba
> Inspecting 10[4](https://github.com/CrunchyData/bridge-cli/actions/runs/3888540565/jobs/6635966303#step:6:5) files
> 
> ...............................................................................F........................
> 
> spec/cb/open_spec.cr:23:24
> [W] Lint/UnusedArgument: Unused argument `args`. If it's necessary, use `_args` as an argument name to indicate that it won't be used.
> > action.open = ->(args : Array(String), _env : Process::Env) do
> ```
> 
> This isn't true though. I've confirmed that the `args` variable is doing what it's supposed to be doing, and if you rename it to `_args`, the build unsurprisingly fails.
> 
> So I think this is a linter bug? What do you guys generally do to work around this kind of stuff?

Yeah, that's odd... I'm looking into to it. Sometimes I've gotten some interesting results. If it comes down to it, we can add an ignore tag on it.

brandur:
@abrightwell Thx. Yeah I figured I'd ask you guys in case it was a known problem before adding the exemption.

brandur:
@abrightwell Mind taking a review pass on this one?

* I'm pretty sure Crystal's linter is buggy — I couldn't find a trivial workaround for that problem, so I ended up replacing it by a stub that just sets a local variable instead of an instance variable. It's not great that this is necessary, but it's the best I could think of.
* Updated the PR description as requested.

abrightwell:
Sorry for the delay on this. I'll go over it this evening.

abrightwell (approved):
lgtm :+1:

brandur:
TY!

PR #94, opened 2023-01-16
Team action improvements.

abrightwell:
Here we're updating the team action to make a few improvements.

* Allow multiple output formats for `list`, `info`, `update`.
* Refactor positional argument for ID to flag `--team`.
* Add examples to help output.
* Change `enforce_sso` to `Bool` from `Bool?`. (API change)
* Update/add new tests.

### DIFF
--- a/src/cb/team.cr
+++ b/src/cb/team.cr
@@ -5,13 +5,53 @@ abstract class CB::TeamAction < CB::APIAction
 
   format_setter format
 
-  private def output_team_details(t : CB::Client::Team)
+  bool_setter show_header, true
+
+  private def output_json(team : CB::Client::Team)
+    output << team.to_pretty_json << '\n'
+  end
+
+  private def output_json(teams : Array(CB::Client::Team))
+    output << {
+      "teams": teams,
+    }.to_pretty_json << '\n'
+  end
+
+  private def output_list(team : CB::Client::Team)
     table = Table::TableBuilder.new(border: :none) do
-      row ["ID:", t.id.colorize.t_id]
-      row ["Name:", t.name.colorize.t_name]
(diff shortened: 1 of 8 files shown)


src/cb/team.cr
```diff
@@ -21,36 +61,61 @@ end
...
     team = client.create_team name
-    output << "Created team #{team}\n"
+
+    case @format
+    when Format::Default
+      output << "Created team #{team}\n"
+    when Format::JSON
+      output_json team
+    else
+      raise_invalid_format @format
```
ngaumont, line 81: This will also allow list and table format

```suggestion
    else
      output(team)
```

src/cb/team.cr
```diff
@@ -21,36 +61,61 @@ end
...
     case @format
     when Format::Default, Format::List
-      output_team_details(team)
+      output_list team
+    when Format::JSON
+      output_json team
+    when Format::Table
+      output_json team
+    else
+      raise_invalid_format @format
```
ngaumont, line 118: There is a typo in the `when  Format::Table` it should use `output_table`
But it could also be written:
```
output(team)
```

src/cb/team.cr
```diff
@@ -83,24 +150,43 @@ class CB::TeamUpdate < CB::TeamAction
...
+    case @format
+    when Format::Default, Format::List
+      output_list team
+    when Format::JSON
+      output_json team
+    when Format::Table
+      output_table team
+    else
+      raise_invalid_format @format
+    end
```
ngaumont, line 162:
```suggestion
    output(team)
```

ngaumont (changes requested):
Nice improvements.
My comments are mainly to add a default ouput method to avoid duplication.

PR #95, opened 2023-02-13
Update `firewall_rules` client.

abrightwell:
Here we're updating the client to use the new
`/networks/:id/firewall-rules` endpoints that are now available in the API.

There are no functional changes to the CLI. This only affects the underlying client.

### DIFF
--- a/src/cb/manage_firewall.cr
+++ b/src/cb/manage_firewall.cr
@@ -2,6 +2,7 @@ class CB::ManageFirewall < CB::APIAction
   Error = Program::Error
 
   eid_setter cluster_id
+  property network_id : String?
   property to_add = [] of String
   property to_remove = [] of String
 
@@ -15,6 +16,7 @@ class CB::ManageFirewall < CB::APIAction
 
   def run
     raise Error.new "--cluster not set" unless cluster_id
+    @network_id = @client.get_cluster(cluster_id).network_id
     remove_all
     add_all
     display_rules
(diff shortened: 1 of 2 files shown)


src/cb/manage_firewall.cr
```diff
@@ -15,14 +16,15 @@ class CB::ManageFirewall < CB::APIAction
 
   def run
     raise Error.new "--cluster not set" unless cluster_id
+    @network_id = @client.get_cluster(cluster_id).network_id
```
abrightwell, line 19: So that we can continue to support the CLI interface as it sits.  There has been some discussion around moving this to be `--network`. However, that has not been finalized yet. So, in the meantime we'll continue to support the command as it sits and just update the underlying client signature and calls.

abrightwell (commented):
@will @ngaumont  Thoughts?

src/cb/manage_firewall.cr
ngaumont: Adding a hidden `--network` would have been nice, but I'm fin with this.
A common pattern, seems to have a `validate` function called inside `run`.
What do you think of adding it now?

abrightwell: Yeah, I actually already have another PR in the works that actually does BOTH of those things. I'm wanting to finalize some of the details of the command first. So, I decided not to include either here in order to keep this changeset as clean as possible.

ngaumont (approved):

abrightwell:
Thanks @ngaumont!

PR #99, opened 2023-03-02
Add `--team` option to `cb list`.

abrightwell:
Here were implementing the `--team` option in `cb list`.  The flag was inadvertently added in #97, but the intent was always to follow up with this feature. So, we go ahead and make it happen here.

Ultimately, the idea is to allow a user to specify an individual team to filter the results. As opposed to always returning all clusters for all teams which they have access to.

Based on #97 until merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 completion.log
 /dist/
 /vendor/
+.envrc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - `cb list` command now accepts `--format`. Supports: `table` and `tree`.
+- `cb list` command now accepts `--team`.
 
 ## [3.2.0] - 2023-02-28
 ### Added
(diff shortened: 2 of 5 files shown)


ngaumont (approved):
I'm fine as-is, but there is weird situation if the user gives a team_id and the team name is used instead:
```
cb list --team 1234 --format=tree
my-awsome-team
└── cluster
```

abrightwell:
> I'm fine as-is, but there is weird situation if the user gives a team_id and the team name is used instead:
> 
> ```
> cb list --team 1234 --format=tree
> my-awsome-team
> └── cluster
> ```

I'm not sure that I follow. What's the expected behavior?

ngaumont:
> > I'm fine as-is, but there is weird situation if the user gives a team_id and the team name is used instead:
> > ```
> > cb list --team 1234 --format=tree
> > my-awsome-team
> > └── cluster
> > ```
> 
> I'm not sure that I follow. What's the expected behavior?

I'm not sure what is the expected behavior, either.
I wanted to point out that the user could enter an eid, but then a name (not eid) is displayed.
The user might not know that `1234` is indeed the eid for the team named `my-awsome-team` and that could cause some confusion.
However, I might be overthinking.

abrightwell:
> I'm not sure what is the expected behavior, either. I wanted to point out that the user could enter an eid, but then a name (not eid) is displayed. The user might not know that `1234` is indeed the eid for the team named `my-awsome-team` and that could cause some confusion. However, I might be overthinking.

AH! I see. Yeah, I could certainly include the `id` as part of the display for sure. That's something I was scratching my head on a little while making these changes. I decided to keep it similar to the `table` output for now. But I'm not opposed to adding more information to the output if it makes sense or reduces confusion for the user.

PR #100, opened 2023-03-02
Remove some `.psqlrc` configurations for `cb psql`.

abrightwell:
Here we're simply removing all but the prompt configuration that is placed in the temporary `.psqlrc` file that the `cb psql` command relies on.  The impetus for this is that some users have reported unexpected behaviors based on what they already have defined in their own `.psqlrc`.  So, in order to respect what they already have configured, we remove any potential behavior configuration from our generated configuration for a psql session.

However, as noted above, we do keep the prompt configuration as that seems like a reasonable configuration to keep for informational purposes and it does not have any unexpected affect on the clients behaviors.


abrightwell:
> Im curious if you have the link, maybe share in slack, why someone that actually wanted a different on error rollback setting. Interactive makes it so it only happens if you're using psql interactively and not in a script. Does someone really want to lose a 5 hour transaction because they typo the next command?

The issue was more related towards the `\x auto` than the rollback. The reasoning why I removed it was simply to not override what a user might have set in their `~/.psqlrc`. Candidly, I have no strong opinion on this one and can see the benefit of keeping it.

will (approved):
👍

PR #101, opened 2023-03-06
Refactor API resource models to `CB::Model` module.

abrightwell:
Here start refactoring API resource models to their own module.

This is a broad refactoring project that I've been wanting to get started.

Currently, all of our API resource models are defined in our `Client`. While I think initially this was the logical place to implement and maintain them, I feel like it would be nice to give them their own namespace.

So that's what I've started doing here.

I'm finding this to be a good exercise as well, as I'm taking the opportunity to reduce redundancy in their implementation as well and updating them to be more in line with what the API currently supports. More specifically, there are some items that have been deprecated by the API as well as being able to merge redundant models such as `Cluster` and `ClusterDetails`. This is because when they were initially implemented the API treated them as separate resources and that is no longer the case.

### DIFF
--- a/spec/cb/cluster_create_spec.cr
+++ b/spec/cb/cluster_create_spec.cr
@@ -115,11 +115,10 @@ Spectator.describe CB::ClusterCreate do
       action.region = "east"
       action.team = team.id
 
-      expect(client).to receive(:create_cluster).and_return(CB::Client::Cluster.new("abc", "def", "my cluster", [] of CB::Client::Cluster))
-
+      expect(client).to receive(:create_cluster).and_return(cluster)
       action.call
 
-      expect(&.output.to_s).to eq "Created cluster abc \"my cluster\"\n"
+      expect(&.output.to_s).to eq "Created cluster pkdpq6yynjgjbps4otxd7il2u4 \"abc\"\n"
     end
   end
 
--- a/spec/cb/completion_spec.cr
+++ b/spec/cb/completion_spec.cr
@@ -1,7 +1,7 @@
 require "../spec_helper"
 
 private class CompletionTestClient < CB::Client
-  def get_clusters(teams : Array(CB::Client::Team)? = nil)
+  def get_clusters(teams : Array(CB::Model::Team)? = nil, flatten : Bool = false)
     [Factory.cluster]
   end
 
--- a/spec/cb/completion_spec.cr
+++ b/spec/cb/completion_spec.cr
@@ -1,12 +1,12 @@
 require "../spec_helper"
 
 private class CompletionTestClient < CB::Client
-  def get_clusters(teams : Array(Team)? = nil, flatten = false)
-    [Cluster.new("abc", "def", "my cluster", [] of Cluster)]
+  def get_clusters(teams : Array(CB::Client::Team)? = nil)
+    [Factory.cluster]
   end
 
   def get_teams
-    [Team.new("def", "my team", false, "manager")]
+    [Factory.team(name: "my team", role: "manager")]
   end
 
   def get_firewall_rules(id)
@@ -89,10 +89,13 @@ end
 Spectator.describe CB::Completion do
   subject(result) { parse(command) }
 
+  let(expected_cluster_suggestion) { ["pkdpq6yynjgjbps4otxd7il2u4\tmy team/abc"] }
+  let(expected_team_suggestion) { ["l2gnkxjv3beifk6abkraerv7de\tmy team"] }
+
   provided command: "cb " { expect(result).to have /^login/ }
 
   sample %w[info rename destroy logs suspend resume].each do |cmd|
-    provided command: "cb #{cmd} " { expect(result).to eq ["abc\tmy team/my cluster"] }
+    provided command: "cb #{cmd} " { expect(result).to eq expected_cluster_suggestion }
     provided command: "cb #{cmd} abc " { expect(result).to be_empty }
   end
 
@@ -106,7 +109,7 @@ Spectator.describe CB::Completion do
     expect(result).to have_option "--version"
 
     result = parse("cb create --fork ")
-    expect(result).to have_option "abc"
+    expect(result).to eq expected_cluster_suggestion
 
     result = parse("cb create --fork abc ")
     expect(result).to have_option "--at"
@@ -120,7 +123,7 @@ Spectator.describe CB::Completion do
     expect(result).to eq([] of String)
 
     result = parse("cb create --replica ")
-    expect(result).to have_option "abc"
+    expect(result).to eq expected_cluster_suggestion
 
     result = parse("cb create --replica abc ")
     expect(result).to have_option "--platform"
@@ -211,10 +214,10 @@ Spectator.describe CB::Completion do
     expect(result).to have_option "512"
 
     result = parse("cb create --team ")
-    expect(result).to have_option "def"
+    expect(result).to eq expected_team_suggestion
 
     result = parse("cb create -t ")
-    expect(result).to have_option "def"
+    expect(result).to eq expected_team_suggestion
 
     result = parse("cb create --ha ")
     expect(result).to have_option "true"
@@ -264,7 +267,7 @@ Spectator.describe CB::Completion do
     expect(result).to have_option "--remove"
 
     result = parse("cb firewall --cluster ")
-    expect(result).to eq ["abc\tmy team/my cluster"]
+    expect(result).to eq expected_cluster_suggestion
 
     result = parse("cb firewall --cluster abc --add 1.2.3/4 --remove 1.2.3.4/5 ")
     expect(result).to_not have_option "--cluster"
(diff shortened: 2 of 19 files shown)


abrightwell:
@ngaumont / @will This is a WIP and I'd like to be able to get all the models refactored in a single PR if possible. But I wanted to put it out there for feedback before I tried to bite off the larger chunk of them.

Ultimately, I think that this will help to organize and clean up the code a little.

Thoughts?

spec/cb/cluster_create_spec.cr
```diff
@@ -115,11 +115,10 @@ Spectator.describe CB::ClusterCreate do
       action.region = "east"
       action.team = team.id
 
-      expect(client).to receive(:create_cluster).and_return(CB::Client::Cluster.new("abc", "def", "my cluster", [] of CB::Client::Cluster))
-
+      expect(client).to receive(:create_cluster).and_return(cluster)
       action.call
 
-      expect(&.output.to_s).to eq "Created cluster abc \"my cluster\"\n"
+      expect(&.output.to_s).to eq "Created cluster pkdpq6yynjgjbps4otxd7il2u4 \"abc\"\n"
```
abrightwell, line 121: Uses the `Factory.cluster` now so needed to update the expected output.

spec/cb/completion_spec.cr
```diff
@@ -89,10 +89,13 @@ end
 Spectator.describe CB::Completion do
   subject(result) { parse(command) }
 
+  let(expected_cluster_suggestion) { ["pkdpq6yynjgjbps4otxd7il2u4\tmy team/abc"] }
+  let(expected_team_suggestion) { ["l2gnkxjv3beifk6abkraerv7de\tmy team"] }
```
abrightwell, line 93: Eventually, I'd like to get this file refactored to make better use of `Spectator` and ultimately make use of `let(cluster) { Factory.cluster }` such that we could make these expected lines be more like:

`let(expected_cluster_suggestion) { ["#{cluster.id}\t#{team.name}/#{cluster.name}"] }`

My thought being that we can eliminate a lot of repetition in the tests, ensure more consistent testing and better ensure realistic outputs.

spec/factory.cr
```diff
@@ -62,22 +62,29 @@ module Factory
...
     }.merge(params)
 
-    CB::Client::ClusterDetail.new **params
+    CB::Model::Cluster.new **params
+  end
+
+  def cluster_status(**params)
+    params = {
+      oldest_backup_at: nil,
+      state:            CB::Model::ClusterStatus::State::Ready,
```
abrightwell, line 84: I don't like the 'length' of this one. But... it is what it is.

src/cb/models/cluster.cr
```diff
@@ -0,0 +1,62 @@
+require "json"
+
+module CB::Model
+  struct Cluster
+    include JSON::Serializable
+
+    property cpu : Float64
```
abrightwell, line 7: I'd like to add some kind of documentation on these. But I'm less interested in trying to keep them up to date if anything changes. Perhaps I'll just include a link to the docs that defines the resource.

src/cb/models/cluster.cr
```diff
@@ -0,0 +1,62 @@
...
+                   @major_version = 0,
+                   @memory = 0,
+                   @network_id = "",
+                   @plan_id = "",
+                   @provider_id = "",
+                   @region_id = "",
+                   @replicas = nil,
+                   @source_cluster_id = nil,
+                   @storage = 0,
+                   @tailscale_active = false)
```
abrightwell, line 59: I wanted to be able to have some kind of `kwarg` support for these. As it would be really helpful for testing, etc.

I'd done some poking around to see what made the most sense and this is what I came up with from looking at the Crystal source for examples. I'm open to suggestions if there is a better way to do this.

src/client/cluster.cr
```diff
@@ -17,39 +15,13 @@ module CB
...
-      oldest_backup : Time?,
-      provider_id : String,
-      network_id : String,
-      region_id : String,
-      maintenance_window_start : Int32?,
-      storage : Int32 do
-      @[JSON::Field(key: "cluster_id")]
-      getter source_cluster_id : String?
+      Array(CB::Model::Cluster).from_json resp.body, root: "clusters"
     end
```
abrightwell, line 25: We can wholesale get ride of this one as it is no longer necessary as there is no longer a separate concept of a `Cluster` and `ClusterDetails` in the API.

src/ext/time.cr
```diff
@@ -0,0 +1,3 @@
+class Time
+  ZERO = Time.utc(1, 1, 1, 0, 0, 0)
```
abrightwell, line 2: I'm still kind of mixed on this one. There are fields from the API resource for `Time` that are not nilable. But I needed a reasonable `default` value for initialization purposes. I suppose it doesn't matter if the property is `Time?` even if the API won't ever return a `null` for the value. But I was trying to keep the model representations as close to what the API would return as possible. I'm open to suggestions here.

src/cb/models/cluster.cr
ngaumont: Agreed, having a top comment with a doc that link either to apiary, or owl might be usefull.

src/ext/time.cr
ngaumont: I'm not sure to understand the purpose of this.
To me, having a default like this is fine. However, you used it only once, so I would directly use the value instead of creating a constant.

ngaumont (commented):
This seems good to me and I think you can go further in the recfactoring process

src/cb/models/cluster.cr
ngaumont: I don't know enough Crystal, yet. Maybe @will has an idea?

abrightwell: Yeah, this is what I was following from the stdlib... https://github.com/crystal-lang/crystal/blob/29f9ac503/src/oauth2/client.cr#L79

src/ext/time.cr
abrightwell: So, the basic idea is that given something like `created_at`. It is should not ever be `nil` from the API. However, for initialization purposes I need to give it some value. I don't want to make the property `Time?` because I feel like that just isn't right.  Effectively, it gives me the ability to give a property a zero value when the value cannot/should not be nilable.

Yeah, currently it's only being referenced once. But should be used more once this refactor continues on to other resource objects.

abrightwell: For something similar in another language like `go`...

`time.Time{}` == 0 sec, 0 nsec.

https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/time/time.go;l=453-457

So, just trying to accomplish the same here.

src/cb/models/operation.cr
```diff
@@ -0,0 +1,66 @@
+require "json"
```
abrightwell, line 1: @ngaumont since this is one that you've been very involved with, did you have thoughts on this approach?

The `Flavor` and `State` are enums from the API. So I figured it made sense to constrain them like that here as well.

I think the interesting things about this approach are the following:

1. Nesting the `enum` in the `Operation` struct. As that makes it quite verbose to utilize, but I'm not that concerned with it given that the only time it's really being utilized directly is in the `specs`.  However, `CB::Model::Operation::State::Ready` does seems kind of cumbersome. But it might be ok.

2. Trying to keep with the naming conventions for enums of `PascalCase` names. This becomes an issue only when outputting the value, both to `io` and `to_json`.  For instance, `HAChange` needs to render as `ha_change`.  So, we have to override `to_s` on the `enum`. I'm ok with that.  As `from_json` works regardless.

3. Along the lines of "2" above. We only need to implement `to_s` to make it possible as that's what `to_json` utilizes.  However, I found that it does not use `to_s(io)` (which is what we should be overriding for general use). So, we implement both.  Seems easy enough and that's why both are there.

ngaumont: I really like the idea of having enum in `cb` as well. I'm just a bit concerned, as this introduce a strong coupling between `cb` and what is returned by the API (adding a new state will brake old cb). I find it okay for `Flavor`, but not for `State`. 

What do you think of keeping the enum for the flavor but stick to bare string for the state ?


TIL `to_json` already use [underscore](https://crystal-lang.org/api/1.7.3/Enum.html#to_json%28json%3AJSON%3A%3ABuilder%29-instance-method), but you still has to update the default `to_s`.


The factory is really helpfull ^^

abrightwell: > I really like the idea of having enum in `cb` as well. I'm just a bit concerned, as this introduce a strong coupling between `cb` and what is returned by the API (adding a new state will brake old cb).

Yes, that's certainly a concern for me.  I'd like to keep it an enum if possible given the API docs specify it as such.  

https://docs.crunchybridge.com/api/cluster-upgrade/#cluster-upgrade-operation

However, your point is very valid. So I'll see if can make it a bit more robust and if I can't, then I'll revert it to a `String`. 👍 

> TIL to_json already use [underscore](https://crystal-lang.org/api/1.7.3/Enum.html#to_json%28json%3AJSON%3A%3ABuilder%29-instance-method), but you still has to update the default to_s.

Yeah, it wasn't necessarily clear to me at first. I'm not 100% on this approach as it should be possible to not have to override `to_s` and instead only provide a `to_s(io)`. But I think we'll need to do some output refactoring in order to make that possible. I'm going to add some `TODO` notes to come back to this and clean it up later if possible.

abrightwell:
@ngaumont I think I'm going to keep this PR limited to the changes it currently proposes. Obviously there are MANY more models to refactor. However, I'm wanting to keep the PR's small enough for meaningful review. Hopefully, I'll get the other done in quick succession.

PR #102, opened 2023-03-15
Refactor `cb maintenance info` output.

abrightwell:
Here we're refactoring how the output for `cb maintenance info` is produced.

Specifically, we're moving the output code out of the `Operation` model and putting it into the action. As well, we make use of the TableBuilder that we are using in other command outputs, using it similar to how we use it with `cb team list`.

Also, as part of this we guarantee the ordering of the output elements. Specifically, we ensure that operations are always first and the `maintenance window` is always last.

Based on #101 until merged.

### DIFF
--- a/src/cb/cluster_upgrade.cr
+++ b/src/cb/cluster_upgrade.cr
@@ -11,32 +11,6 @@ abstract class CB::Upgrade < CB::APIAction
       missing << "cluster" unless cluster_id
     end
   end
-
-  def display_operations(c_id, operations, maintenance_only : Bool)
-    details = {
-      "maintenance window" => MaintenanceWindow.new(c_id.maintenance_window_start).explain,
-    }
-
-    operation_kind = "operations"
-    if maintenance_only
-      operation_kind = "maintenance operations"
-      operations = operations.select { |op| op.flavor != CB::Model::Operation::Flavor::HAChange }
-    end
-
-    operations.each do |op|
-      details[op.flavor.to_s] = op.one_line_state_display
-    end
-
-    if operations.empty?
-      output << "  no #{operation_kind} in progress\n".colorize.bold
-    end
-
-    pad = (details.keys.map(&.size).max || 8) + 2
-    details.each do |k, v|
-      output << k.rjust(pad).colorize.bold << ": "
-      output << v << '\n'
-    end
-  end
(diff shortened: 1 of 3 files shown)


src/cb/cluster_upgrade.cr
```diff
@@ -11,32 +11,6 @@ abstract class CB::Upgrade < CB::APIAction
...
-    if operations.empty?
-      output << "  no #{operation_kind} in progress\n".colorize.bold
-    end
-
-    pad = (details.keys.map(&.size).max || 8) + 2
-    details.each do |k, v|
-      output << k.rjust(pad).colorize.bold << ": "
-      output << v << '\n'
-    end
-  end
```
abrightwell, line 39: Just pushed this down into the `UpgradeStatus` actions as that's the only action that utilizes it.

src/cb/cluster_upgrade.cr
```diff
@@ -118,6 +92,28 @@ class CB::UpgradeStatus < CB::Upgrade
...
+
+    output_default(operations, c.maintenance_window_start)
+  end
+
+  def output_default(operations, maintenance_window)
+    table = Table::TableBuilder.new(border: :none) do
+      if !operations.empty?
+        operations.each do |op|
+          from = " (Starting from: #{op.starting_from})" if op.starting_from
+          row ["#{op.flavor.colorize.bold}:", "#{op.state}#{from}"]
```
abrightwell, line 108: This is where we add back the function of `one_line_state_display`

src/cb/models/operation.cr
```diff
@@ -11,12 +11,8 @@ module CB::Model
...
       Resize
 
-      def to_s
-        super.to_s.underscore
-      end
-
-      def to_s(io : IO) : Nil
-        io << to_s
+      def to_s(io : IO)
+        io << to_s.underscore.gsub('_', ' ')
```
abrightwell, line 15: This is the cleanup to this that I mentioned in #101.

The idea is that we are basically able to take a value like `HAChange` and output it as `ha change`.

abrightwell:
@ngaumont I think this one is a good step forward, there are still some other clean up items that are related that I want to address, but intend to follow up on some of that in a later PR.  Thoughts?

PR #103, opened 2023-03-16
More model refactoring

abrightwell:
Here we're wrapping up the high level `CB::Model` refactor by moving all of the remaining and relevant models out of `CB::Client` into `CB::Model`.  Much of this is simply a copy/paste of the structs. However, we also take the opportunity to add missing `Factory` methods as well as updating related specs to use the `Factory` over direct instantiation of the models.

### DIFF
--- a/spec/factory.cr
+++ b/spec/factory.cr
@@ -117,7 +117,7 @@ module Factory
       team_id:     "l2gnkxjv3beifk6abkraerv7de",
     }.merge(params)
 
-    CB::Client::Network.new **params
+    CB::Model::Network.new **params
   end
 
   def operation(**params)
--- a/spec/factory.cr
+++ b/spec/factory.cr
@@ -202,7 +202,7 @@ module Factory
       email:      "test@example.com",
     }.merge(params)
 
-    CB::Client::TeamMember.new **params
+    CB::Model::TeamMember.new **params
   end
 
   def tempkey(**params)
--- a/spec/factory.cr
+++ b/spec/factory.cr
@@ -20,7 +20,7 @@ module Factory
       size_bytes:  123.to_u64,
     }.merge(params)
 
-    CB::Client::Backup.new **params
+    CB::Model::Backup.new **params
   end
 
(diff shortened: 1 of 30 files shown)


spec/factory.cr
```diff
@@ -110,25 +130,54 @@ module Factory
...
+  def role_user(**params)
+    params = {
+      account_email: "user@example.com",
+      name:          "u_mijrfkkuqvhernzfqcbqf7b6me",
+      password:      "secret",
+      uri:           URI.parse "postgres://u_mijrfkkuqvhernzfqcbqf7b6me:secret@example.com:5432/postgres",
+    }.merge(params)
+
+    CB::Model::Role.new **params
   end
```
abrightwell, line 181: I renamed these from to `role_*` to keep them 'grouped' together as this file potentially grows. Also, I much prefer the ordering of the name being based from general -> specific.  As it helps with code suggestions, etc.

spec/factory.cr
```diff
@@ -110,25 +130,54 @@ module Factory
...
+      id:           "jkon7qbrzzccrgwn346w3niloe",
+      display_name: "Example Plan",
     }.merge(params)
 
-    CB::Client::Role.new **params
+    CB::Model::Plan.new **params
   end
 
-  def system_role(**params)
+  def provider(**params)
```
ngaumont, line 142: Just a comment, not a suggested modification.

In the future, this could be customized to allow a call like this, if needed:
```ruby
Factory.provider(
  id: "gcp",
  display_name: "GCP",
  plans: [{id: "standard-128", display_name: "Standard-128"}],
  regions: [{id: "us-central1", display_name: "US Central 1", location: "Iowa"}]
```

ngaumont (approved):

spec/factory.cr
abrightwell: I completely agree. I had considered doing that but figured I'd come back through later to do a complete pass on ALL of the methods. 👍

abrightwell:
Thanks!

PR #104, opened 2023-03-16
Move `models` directory up a level.

abrightwell:
Here just moving the models directly up a level out of `./src/cb` to `./src`.  This is purely for better orgnaization.

Based on #103 until merged.


ngaumont (approved):

abrightwell:
Thanks!

PR #105, opened 2023-03-20
Add flake for devShells

will:
This flake adds dev shells for x86 linux and both x86/arm mac development. Right now it's using a crystal flake I made in my own github account, but we can move that to crunchy's nixpkgs once we get that going.

The flake does not provide a build yet for cb, but that should be possible in the future.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,29 @@
+{
+  description = "CrunchyBridge CLI";
+
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs-crystal = {
+      url = "github:will/nixpkgs-crystal";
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,31 @@
+{
+  description = "CrunchyBridge CLI";
+
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
```
will, line 5: I went back from my earlier suggestion of inlining this I made in the other repo. I'm starting to look into some other things that help building flakes and might switch to them. I figure use the import linked like this while we're figuring things out, then maybe solidify down later

abrightwell: Cool, that WFM. 👍

abrightwell (approved):
LGTM. 👍

PR #108, opened 2023-04-04
Add nix default package

will:
This changes the build version to remove the date and instead have a git sha, which makes the builds more reproducible.

It also add a default package so you can use cb like:

    $ nix run . -- version
    $ nix run github:crunchydata/bridge-cli\?ref=cafee81 -- version

and after this is merged, just

    $ nix run github:crunchydata/bridge_cli -- version
    $ nix shell github:crunchydata/bridge_cli

etc

Right now the version of crystal the flake this is using is 1.8.0-dev. ~My impression is that~ 1.8.0 will be released ~soon~ on April 12th, and we can switch to that. However were planning on stamping a release version anytime soon, we should probably have that version using crystal 1.7.3 proper

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ completion.log
 /dist/
 /vendor/
 .envrc
+result
--- a/flake.lock
+++ b/flake.lock
@@ -32,11 +32,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1679281263,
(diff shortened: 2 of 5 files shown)


abrightwell (approved):
WFM 👍 

Also, I don't foresee a release prior to April 12, unless something critical comes up.  So, I'd just keep with 1.8.0 for now.

PR #113, opened 2023-04-14
Use updated packaging for crystal

will:
I've made a lot of improvements to the crystal flake I've been working on.

* A function that applies a bunch of defaults to make a crystal package, to remove a bunch of the boilerplate there.
* An "extra wrapped" crystal compiler.

I don't know if there is a more idiomatic nix name for doing this. I was already wrapping the crystal compiler with all the env vars it needs to find llvm, know about openssl, and find the other libraries like pcre2, bohem-gc, etc. However for projects that needed extra libraries, like libssh2, you would need to specify them both in the devshell, as well as when making a package for your program. On top of that, since the compiler wasn't wrapped with these new libraries, `crystal i` wouldn't know about them. It was possible to override the entire crystal derivation with extra buildInputs, but then you'd have to recompile the compiler, which seemed like a waste. So I added a new derivation for crystal that uses the original one as input (which comes wrapped once with the base libraries), and then wrap it again with the extra libraries you need. The end result is that now you only need to say libssh2 in one place, and `crystal i` works. I think it's neat.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1678901627,
-        "narHash": "sha256-U02riOqrKKzwjsxc/400XnElV+UtPUQWpANPlyazjH0=",
+        "lastModified": 1681202837,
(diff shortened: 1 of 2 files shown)


abrightwell (approved):
SGTM. 👍

abrightwell:
💥 Works! 👍

abrightwell (approved):

PR #114, opened 2023-04-20
fix: shorter time period for starting_from

ngaumont:
Following
https://crunchydata.slack.com/archives/C012GUR4XAT/p1681809181226709, the starting_from parameter should be reduced from 7 days to 72 hours. This is needed as the server is created right away, but not billed to the user.

I don't know if something should be added in the changelog

### DIFF
--- a/src/cb/cluster_upgrade.cr
+++ b/src/cb/cluster_upgrade.cr
@@ -28,10 +28,10 @@ abstract class CB::UpgradeAction < CB::Upgrade
     raise Error.new "Must use '--starting-from' or '--now' but not both." if starting_from && now
 
     if now
-      starting_from = Time.utc
+      @starting_from = Time.utc
     end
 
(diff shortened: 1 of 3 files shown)


src/cb/cluster_upgrade.cr
```diff
@@ -31,7 +31,7 @@ abstract class CB::UpgradeAction < CB::Upgrade
       starting_from = Time.utc
     end
 
-    raise Error.new "'--starting-from' should be in less than a week" if (start = starting_from) && start > (Time.utc + Time::Span.new(days: 7))
+    raise Error.new "'--starting-from' should be in less than 72 hours" if (start = @starting_from) && start > (Time.utc + Time::Span.new(hours: 72))
```
ngaumont, line 34: Here, if I don't add `@` the test failed and I it find strange.
I can confirm that without it, the command does not validate starting-from

abrightwell (approved):
LGTM 👍

PR #115, opened 2023-04-21
Partially revert 8dd4b5d.

abrightwell:
Here we're partially reverting part of 8dd4b5d. Specifically, we are removing the `LIBRARY_PATH` addition that was made before we had made more progress towards integrating the use of nix dev shells via `flake.nix`.  We've since resolved the issue and it is no longer necessary to include this ENV when running the specs on MacOS.


ngaumont (approved):

abrightwell:
Thanks!

PR #116, opened 2023-04-26
Cross platform forks and replica.

abrightwell:
Here we're updating `cb create --fork` and `cb create --replica` to
support cross platform creation. Previously, this was not possible as
the fork and replica client requests were always defaulting to the
target/primary cluster's platform, region and plan. Now, it is possible
to create a fork or replica in another platform in the same way it is
that a standalone cluster is created.

For instance, to create an AWS based fork in GCP:

```
$ cb create --fork <ID> --platform gcp --region uswest-1 --plan
standard-8
```

If `--name` is omitted, then the value will default to `Fork of <name>`
where `<name>` is the name of the source cluster.  The same is true for
replicas, expept that it will `Replica of <name>`.

This also allows for the creation of forks and replicas across regions
within the same platform as well. Though, it still requires the full
qualification of specifying the target `--platform` and `--plan`.
Future improvements on this might be possible to simply allow the user
to only specify the `--region`. However, for now it seems prudent to
simply keep this current requirement as it helps to ensure (through
completion) that the any region or plan specified is correct/available
for a specific platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and a pending upgrade respectively.
 
 ### Fixed
+- `cb destroy` completion to include `--confirm`.
 - `cb psql` no longer overrides a users `.psqlrc` with `\x auto` which was
   causing unexpected formatting for some users.
 - `cb info` now returns new cluster states: `resuming`, `suspended`, `suspending`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `cb create --fork` and `cb create --replica` now supports cross platform
+  creation.
 - `cb list` command now accepts `--format`. Supports: `table` and `tree`.
 - `cb list` command now accepts `--team`.
 - `cb maintenance update` and `cb upgrade update` update a pending maintenance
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -6,18 +6,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - `cb list` command now accepts `--format`. Supports: `table` and `tree`.
 - `cb list` command now accepts `--team`.
 - `cb maintenance update` and `cb upgrade update` update a pending maintenance
   and a pending upgrade respectively.
 
 ### Fixed
+- `cb destroy` completion to include `--confirm`.
+- `cb info` now returns new cluster states: `resuming`, `suspended`, `suspending`.
+- `cb logs` fails without error message.
+- `cb maintenance cancel` now only cancels only maintenances.
```
ngaumont, line 20: Another typo from me… Could you also fix this one ?
```suggestion
- `cb maintenance cancel` now cancels only maintenances.
```

src/cb/cluster_create.cr
```diff
@@ -15,60 +15,69 @@ class CB::ClusterCreate < CB::APIAction
...
+                  name: @name.to_s,
+                  plan_id: @plan.to_s,
+                  provider_id: @platform.to_s,
+                  postgres_version_id: @postgres_version,
+                  region_id: @region.to_s,
+                  storage: @storage,
+                  team_id: @team.to_s,
+                  network_id: @network,
+                )
+              end
```
ngaumont, line 76: I find this more readable, but let you decide:

```suggestion
    param = {
      name: @name.to_s,
      network_id: @network,
      plan_id: @plan,
      provider_id: @platform,
      region_id: @region,
     }
    cluster = if fork
                @client.create_fork CB::Client::ForkCreateParams.new(**param.merge(
                  cluster_id: @fork.to_s,
                  is_ha: @ha,
                  storage: @storage,
                  target_time: @at,
                ))
              elsif replica
                @client.create_replica CB::Client::ReplicaCreateParams.new(**param.merge(
                  cluster_id: @replica.to_s,
                ))
              else
                @client.create_cluster CB::Client::ClusterCreateParams.new(**param.merge(
                  is_ha: @ha,
                  postgres_version_id: @postgres_version,
                  storage: @storage,
                  team_id: @team.to_s,
                ))
              end
```

If you keep your implementation, could you sort the parameters (`network_id`). It will help identify the common part.

ngaumont (approved):

src/cb/cluster_create.cr
abrightwell: Yeah, I'm cool with this approach.

PR #117, opened 2023-04-26
Use nix for checks and CI

will:
This adds two things. First a single command `check` that runs
    formatting checks, ameba checks, and the specs not only in parallel,
    which is cool enough, but also each check knows what files it depends on
    and wont run if those haven't changed. So if you run `check` and then
    edit the readme, it wont run the specs again.

  Second it uses these checks on CI and also caches the check builds
    between pull requests. So also on CI we can have all the checks only run
    if they need to. And all of this without needing to do ever more careful
    github caching and filters in yaml and so on. Just a single job that
    takes care of everything.


will:
ok first ci build with nothing cached
image
going to push a commit with just a readme change to see what happens

will:
Even after I rebased everything down to a single commit, it still knew that nothing really changed: 
image

.ameba.yml
```diff
@@ -1,15 +1,65 @@
+# This configuration file was generated by `ameba --gen-config`
+# on 2023-04-25 14:12:48 UTC using Ameba version 1.4.2.
+# The point is for the user to remove these configuration records
+# one by one as the reported problems are removed from the code base.
```
will, line 4: The update to ameba caused some new breaking, so I figured put this in for now, and we can work later to remove these.

abrightwell: Cool, yeah, I have a commit in #116 that attempts to somewhat resolve most of these, I think. I'd be curious to see if merging with that resolves these as well (assuming that they should).

abrightwell (approved):

.ameba.yml
will: I rebased and recreated this list

abrightwell: Cool, I'll put it on my TODO to come back later and attempt to resolve what I can with these.

will:
Just fyi @abrightwell I'm sometimes seeing an error `Error setting time on file: '/tmp/.cache/crystal/nix-store-d28wp3pfvnm41smgv85a6wx0f055d7vj-crystal-1.8.1-lib-crystal-ecr-process.cr': Operation not permitted (File::Error)` when doing the specs. Running a second time "fixes" it. I think some changes need to be done to point the build cache to some other directory maybe

abrightwell:
> Just fyi @abrightwell I'm sometimes seeing an error `Error setting time on file: '/tmp/.cache/crystal/nix-store-d28wp3pfvnm41smgv85a6wx0f055d7vj-crystal-1.8.1-lib-crystal-ecr-process.cr': Operation not permitted (File::Error)` when doing the specs. Running a second time "fixes" it. I think some changes need to be done to point the build cache to some other directory maybe

Sounds good. Thanks for the heads up. 👍

PR #118, opened 2023-04-26
Update `flake.nix`.

abrightwell:
Updates flake.nix to incorporate upstream changes to `crunchy-nixpkgs`. As well, we add `ameba` to the `devShell`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,8 +16,9 @@
         pkgs = nixpkgs.legacyPackages.${system};
         crystal-pkgs = nixpkgs-crunchy.packages.${system};
 
+        ameba = crystal-pkgs.ameba;
         crystal = crystal-pkgs.crystal;
-        crystalWrapped = crystal-pkgs.extraWrapped.override {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -16,8 +16,9 @@
         pkgs = nixpkgs.legacyPackages.${system};
         crystal-pkgs = nixpkgs-crunchy.packages.${system};
 
+        ameba = crystal-pkgs.ameba;
         crystal = crystal-pkgs.crystal;
-        crystalWrapped = crystal-pkgs.extraWrapped.override {
+        crystalWrapped = crystal-pkgs.crystalWrapped.override {
```
abrightwell, line 21: I'm not sure if this was an intended change upstream, but seemed necessary to ensure things still worked here.

will: It was, since eventually there'll be non-crystal packages there and it wouldn't make sense without having the word crystal in the name

will (approved):

abrightwell:
Thanks!

PR #119, opened 2023-05-09
Fix confirmation issue with `upgrade` and `maintenance` update.

abrightwell:


abrightwell:
@ngaumont / @will this is a just a quick fix on this one. I'm going to come back and do some clean up around this and add some more/better tests to account for it in a follow up PR. But I wanted to get this in for now as a quick fix and release update.

abrightwell:
Also, FWIW, the reason I didn't update the tests in this PR is that I need to figure out the best way to handle 'input' with Spectator. In RSpec, I believe you can do something like `allow($stdin).to receive...` but it's less clear how to do the equivalent with Spectator. Mock, Double, etc... sure... but I have a TODO to poke at it more not only for this feature but others as well. Just FYI.

ngaumont (approved):

abrightwell:
Thanks!

PR #122, opened 2023-05-11
Use the ha dedicated endpoint

ngaumont:
This also prevents the following usage:

```
cb upgrade startr --ha --storage 150
```

This needs https://github.com/CrunchyData/priv-all-cloud-owl/pull/2234 to be deployed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- High availability changes with `cb upgrade start` must be made without any other changes to the cluster.
+
 ## [3.4.0] - 2023-08-14
 ### Added
 - `cb config-param` command to manage supported cluster configuration
--- a/spec/cb/cluster_upgrade_spec.cr
+++ b/spec/cb/cluster_upgrade_spec.cr
@@ -22,18 +22,143 @@ Spectator.describe CB::UpgradeStart do
         action.validate
       }.to raise_error Program::Error, "'--starting-from' should be in less than 72 hours"
     end
+
+    it "errors if both ha and any other upgrade is set at the same time" do
+      action.cluster_id = cluster.id
+      action.ha = false
+      action.storage = 150
+
+      expect {
+        action.validate
+      }.to raise_error Program::Error, "--ha is not valid with any other modifications such as '--storage' or '--version'"
+    end
+  end
+  describe "#run" do
+    it "prints cluster upgrade started" do
+      action.cluster_id = cluster.id
+      action.confirmed = true
+
+      expect(client).to receive(:get_cluster).and_return(cluster)
+      expect(client).to receive(:upgrade_cluster).and_return([] of CB::Model::Operation)
+
+      action.call
+
+      expect(&.output.to_s).to eq "  Cluster #{action.cluster_id} upgrade started.\n"
+    end
+  end
+end
+
+Spectator.describe CB::UpgradeStart do
+  subject(action) { described_class.new client: client, output: IO::Memory.new }
+
+  mock_client do
+    def enable_ha(id : Identifier)
+      Factory.operation(
+        flavor: CB::Model::Operation::Flavor::HAChange,
+        state: CB::Model::Operation::State::EnablingHA
+      )
+    end
   end
 
-  it "#run prints cluster upgrade started" do
+  let(cluster) { Factory.cluster }
+
+  it "uses action/enable-ha to enable" do
     action.cluster_id = cluster.id
     action.confirmed = true
+    action.ha = true
 
     expect(client).to receive(:get_cluster).and_return(cluster)
-    expect(client).to receive(:upgrade_cluster).and_return([] of CB::Model::Operation)
 
     action.call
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- High availability change with `cb upgrade start`  should be done without any other upgrades.
```
abrightwell, line 9:
```suggestion
- High availability changes with `cb upgrade start` must be made without any other changes to the cluster.
```

abrightwell (commented):
Just a few comments.  Also, ensure we update the `completions` to not suggest `--ha` if any of the others are present. And vice-versa.

src/client/cluster.cr
```diff
@@ -198,6 +198,22 @@ module CB
       CB::Model::Cluster.from_json resp.body
     end
 
+    # https://crunchybridgeapi.docs.apiary.io/#reference/0/clustersclusteridactionsdisable-ha/disable-high-availability
+    def disable_ha(id : Identifier) : CB::Model::Operation?
+      resp = put "clusters/#{id}/actions/disable-ha"
+      json = JSON.parse(resp.body)
+      return CB::Model::Operation.from_json resp.body if (flavor = json["flavor"]?) && !flavor.as_s.empty?
```
ngaumont, line 205: This is a rather ugly work around the endpoint returning an empty operation instead of nothing.

spec/cb/cluster_upgrade_spec.cr
```diff
@@ -22,18 +22,143 @@ Spectator.describe CB::UpgradeStart do
...
+      expect(client).to receive(:upgrade_cluster).and_return([] of CB::Model::Operation)
+
+      action.call
+
+      expect(&.output.to_s).to eq "  Cluster #{action.cluster_id} upgrade started.\n"
+    end
   end
+end
+
+Spectator.describe CB::UpgradeStart do
```
ngaumont, line 51: This whole test is messed up, but this is the solution that worked at the end.

I first try to just `expect(client).to receive(:enable_ha).and_return(Factory.operation)`, but `nil` was returned instead of the operation…
I spent a few hours on this, but without any success.

If you have a better solution, please tell me

src/cb/cluster_upgrade.cr
```diff
@@ -43,9 +45,31 @@ class CB::UpgradeStart < CB::UpgradeAction
...
+        wanted = "disabled"
+        operation = client.disable_ha(Identifier.new c.id)
+      end
 
-    client.upgrade_cluster self
-    output.puts "  Cluster #{c.id.colorize.t_id} upgrade started."
+      case operation.try &.state
+      when nil
+        output.puts "  High availability already #{wanted} on cluster #{c.id.colorize.t_id}."
+      when CB::Model::Operation::State::DisablingHA
```
abrightwell, line 62:
```suggestion
      when nil
        wanted = ha ? "enabled" : "disabled"
        output << "  High availability already #{wanted} on cluster #{c.id.colorize.t_id}."
      when CB::Model::Operation::State::DisablingHA
```

src/cb/cluster_upgrade.cr
```diff
@@ -43,9 +45,31 @@ class CB::UpgradeStart < CB::UpgradeAction
...
+        output.puts "  Disabling high availability on cluster #{c.id.colorize.t_id}."
+      when CB::Model::Operation::State::EnablingHA, CB::Model::Operation::State::WaitingForHAStandby
+        output.puts "  Enabling high availability on cluster #{c.id.colorize.t_id}."
+      else
+        output.puts "  Operation not recognized: #{operation.inspect}"
+      end
+    else
+      client.upgrade_cluster self
+      output.puts "  Cluster #{c.id.colorize.t_id} upgrade started."
+    end
```
abrightwell, line 72:
```suggestion
        output << "  Disabling high availability on cluster #{c.id.colorize.t_id}."
      when CB::Model::Operation::State::EnablingHA, CB::Model::Operation::State::WaitingForHAStandby
        output << "  Enabling high availability on cluster #{c.id.colorize.t_id}."
      else
        output << "  Operation not recognized: #{operation.inspect}"
      end
    else
      client.upgrade_cluster self
      output << "  Cluster #{c.id.colorize.t_id} upgrade started."
    end
```

src/client/cluster.cr
abrightwell: I'm not sure I follow. I think I would expect that the response would never be empty. 🤔 

For instance, I could understand that HA might already be disabled, but returning an 'operation' that is representative of that might seem appropriate.

Something like:

```
{
  "flavor": "ha_change",
  "starting_from": null,
  "state": "ready"
}
```

Basically acknowledging that the state change was requested, BUT it's already at that state... so, it's 'ready'. Otherwise, I'd expect some kind of error. Though, I don't think an error is appropriate if the request was successful but the cluster was already in the desired state. Thoughts?

ngaumont:
@abrightwell the owl PR is now deployed. This is the error showing up when disabling HA on a cluster that already does not have HA:
```
cb upgrade start --cluster 6zysop5j2zhuxdgwrcp4batnmq --ha false --confirm
error: 400 Bad Request
       PUT to /clusters/6zysop5j2zhuxdgwrcp4batnmq/actions/disable-ha
       code: 
       is_transient: false
       message: High availability is already disabled.
       request_id: e6d36d96-a400-4e55-9aa4-59c513f39497
```

PR #128, opened 2023-05-26
Refactor `cb login` to support browser login.

abrightwell:
This is a major change to how the login flow works as well as how API
requests are authenticated.

Previously, `cb login` would prompt the user for a `client_id` and a
`client_secret`. Later it made the `client_id` optional given changes to
the format of the API key. The command would persist this key/secret and
then use it to procure access tokens when it was ready to start making
requests. Unfortunately this wasn't ideal as it 1) wasn't clear if a
login was successful and 2) logging out was confusing given the
relationship of the API key to access tokens. In summary, while it
served it's purpose this flow was no ideal in the long run. So we're
making it better.

Now, when running `cb login` the user will be prompted to open a browser
by pressing 'any key'.  Or they can abort the process by pressing `q`.
This is a pretty standard prompt so should be familiar to most. The
browser will open to a CLI login specific page at crunchybridge.com.
Here they will enter their credentials in the same way they would if
they were logging into the dashboard. Upon successful login, the user
can return to the CLI. Once the CLI has received notification that the
login attempt was successful, it will notify the user.

The process should result in something similar to the below example:

```
❯ cb login
Press any key to open a browser to login or q to exit:
Waiting for login... done
Logged in as user@example.com
```

Since it's likely that some users of `cb` require headless or scripted
interactions. We've added support for `CB_API_KEY` env. If this is
present then it will take precedence over any other credential that
exists on the system. This value is NOT stored by `cb` and so it will be
on the user to set the env appropriately or passing it along with calls.

For example:

```
❯ CB_API_KEY=cbkey_<secret> cb list
```

It is important to note the `cb login` command will not work if this env
is set. In fact, it will error accordingly.

```
❯ CB_API_KEY=cbkey_secret cb login
error: Cannot login with CB_API_KEY set.
```

Also as a result of these changes `cb token` will only work if an API
key is present.  This is not necessarily a restriction of `cb` but the
API itself as it will not issue access tokens unless the request to do
so is do so with an API key. This might change in the future, but for
now it is something that `cb` has to account for. Similar to the above
with `cb login`, `cb token` will notify the user appopriately if
attempting to procure an access token by any other means.  Generally, it
seems that this restriction for now is fine as most use of this command
are likely for scripting purposes only and an API key is likely already
in use.

This is just the first step in improving this flow. Followup efforts
will include enabling `keyring` support that will store the credential
in system supported security vaults. `Keychain` on MacOS and `GNOME
Keyring` on Linux. If neither of these two systems are supported on the
target OS, then we'll default back to the current `XDG_CONFIG_HOME`
approach. Though it's likely this will be rare at best.


flake.nix
```diff
@@ -25,6 +25,11 @@
         src = filterSrc (shardFiles ++ [ "Readme" "Changelog" ]);
         specSrc = filterSrc shardFiles;
         lintSrc = filterSrc [ ".ameba.yml" ];
+
+        darwinBuildInputs = [
+          pkgs.darwin.apple_sdk.frameworks.Foundation
+          pkgs.darwin.apple_sdk.frameworks.Security
```
abrightwell, line 31: These will be necessary when we fully bring in the keyring support. I can remove from this PR for the time being if desired.

shard.yml
```diff
@@ -2,22 +2,28 @@ name: cb
...
   ssh2:
     github: spider-gazelle/ssh2.cr
-  promise:
-    github: spider-gazelle/promise
   tallboy:
     github: epoch/tallboy
 
 development_dependencies:
-  ameba:
-    github: crystal-ameba/ameba
```
abrightwell, line 18: Removed as it's included along with the flake devshell.

src/cb/cacheable.cr
```diff
@@ -1,5 +1,5 @@
 require "json"
```
abrightwell, line 1: Might be able to do away with this one... but I figured I'd evaluate and address it in a followup.

src/cb/creds.cr
```diff
@@ -1,40 +1,82 @@
...
+# require "keyring"
 require "./dirs"
 
-struct CB::Creds
-  include JSON::Serializable
+# TODO(abrightwell): Below we've intentionally commented out the path that will
+# persist the credentials to the keyring service. This is so that we can work
+# through and ensure the login flow is solid before introducing the new storage
+# of credentials.  This WILL be updated before the next official release of
+# `cb`.
```
abrightwell, line 12: Mostly, things work the same as before. Just setting things up to use keychain, etc. I've tested it locally and it all works. But we need to resolve some release related items first before incorporating this such that the end user experience of working with keychain is effectively seamless or otherwise goes unnoticed.

src/cb/psql.cr
```diff
@@ -62,7 +63,7 @@ module CB
     end
 
     private def ensure_cert(team_id) : String
-      cert_dir = CB::Creds::CONFIG / "certs"
+      cert_dir = CB::Dirs::CERTS_DIR
```
abrightwell, line 66: Just a little cleanup.

src/cb/whoami.cr
```diff
@@ -2,8 +2,10 @@ require "./action"
 
 class CB::WhoAmI < CB::APIAction
   def run
-    output << "user id: ".colorize.t_id << client.token.user_id << "\n"
-    output << "   name: ".colorize.t_id << client.token.name << "\n"
-    output << "   host: ".colorize.t_id << client.host << "\n"
+    account = client.get_account
```
abrightwell, line 5: Actually pull the information from the account API. Could also perhaps pull it from the credentials metadata. But not all of it exists there yet. We could certainly add it though before the next release. It made sense, however, to check the API for the most up-to-date information.

src/cb/creds.cr
```diff
@@ -1,40 +1,83 @@
...
+      #   # Let's remove any old credentials we might have stored. This is so that
+      #   # storing credentials behaves more like an 'update' than a 'create'
+      #   # since if an entry already exists it will raise an error.
+      #   Keyring.delete(host, account) rescue Keychain::ItemNotFound
+      #   stored = Keyring.set(host, account, secret)
+      # rescue Keyring::NotAvailable
+      #   creds.secret = secret
+      # end
+
+      stored = File.open(Dirs::CONFIG/host, "w", perm: 0o600) do |f|
```
ngaumont, line 43: Do you still want to create the directory ? 
```suggestion
      Dir.mkdir_p CONFIG
      stored = File.open(Dirs::CONFIG/host, "w", perm: 0o600) do |f|
```

src/cb/dirs.cr
```diff
@@ -4,7 +4,8 @@ module CB::Dirs
   # $XDG_CONFIG_HOME defines the base directory relative to which user-specific
   # configuration files should be stored. If $XDG_CONFIG_HOME is either not set
   # or empty, a default equal to $HOME/.config should be used.
-  CONFIG = cb_path "XDG_CONFIG_HOME", ".config"
+  CONFIG    = cb_path "XDG_CONFIG_HOME", ".config"
+  CERTS_DIR = Path[CONFIG, "certs"]
```
ngaumont, line 8: You might want to do it for `host` also

ngaumont (changes requested):
This seems pretty good and work on my machine.
I have only very minor comment/question.

src/cb/login.cr
```diff
@@ -1,37 +1,71 @@
...
-    output.puts "add credentials for #{host.colorize.t_name} #{hint}"
+    property client : CB::Client = CB::Client.new CB::HOST
 
-    output.print "  application secret: "
-    secret = input.noecho { input.gets }
+    def open_browser? : Bool
+      output << "Press any key to open a browser to login or "
+      output << "q".colorize.yellow
+      output << " to exit: "
+      response = input.read_char
```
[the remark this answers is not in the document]
abrightwell: I'm not sure I follow, what do you mean exactly?

src/cb/login.cr
```diff
@@ -1,37 +1,71 @@
...
-    creds = Creds.new(host, secret).store
-    status = creds ? "OK".colorize.green.bold : "Failed".colorize.red.bold
-    output << "#{status}\n"
-    creds
+      si_get_params = Client::SessionIntentGetParams.new(session_intent_id: session_intent.id, secret: "#{session_intent.secret}")
+      Retriable.retry(on: NoSession, base_interval: 1.seconds, multiplier: 1.0, rand_factor: 0.0) do
+        session_intent = @client.get_session_intent(si_get_params)
+        raise Error.new "Timeout" if Time.utc > session_intent.expires_at
+        raise NoSession.new unless session_intent.session
+      end
```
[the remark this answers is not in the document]
abrightwell: Gotcha... how long did you wait? The request rate SHOULD be within the allowed limit. Though, I've seen this happen to me once before.

src/cb/creds.cr
abrightwell: Yeah, we certainly should... good catch. 👍

src/cb/dirs.cr
ngaumont: Yes, that's what I meant, but do as you want.

src/cb/login.cr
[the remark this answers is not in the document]
ngaumont: On heroku, I can type one key, like `y`.
On cb, I have to type `y` (first key) and then press enter (second key).

src/cb/login.cr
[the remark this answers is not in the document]
ngaumont: I waited something like 40~50 second.

Continuing the `heroku` and `cb` comparison, `heroku` displays the opened url: `Opening browser to …`.
I don't have an opinion on whether it should be displayed or not in `cb`.

src/cb/login.cr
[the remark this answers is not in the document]
abrightwell: Ah! Yeah, this should be how it works. I thought it did, but all my manual testing I've just being pressing `enter`. 🤦‍♂️ I will update, as what you describe/expected is what it should be.

src/cb/dirs.cr
abrightwell: Not a bad idea... let me think on it. I don't like having to pass the host all over, so this could be a good solution.

src/cb/login.cr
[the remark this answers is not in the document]
abrightwell: I'll be looking into this on the API side as it should be 429'ing. However, I have improved the 'experience' to some degree. It COULD be better. But that's going to require more extensive improvements in other ways. But, will need to follow up in a different PR.

src/cb/dirs.cr
abrightwell: I'm going to come back to this one. I like the idea on centralizing it, but since `CONFIG/host` is a file not a dir, I don't think this is the right place for it. I've giving myself a TODO for it, so I'll come back to this.

ngaumont (approved):
This seems good to me.

I have a last minor comment.
On `cb token`, I rightfully get the following error:
```
You're using an invalid API key. Please procure a new one via the dashboard to continue.
```
Adding `CB_API_kEY=` works as expected.
The wording could however be a little bit improved.
Something like: `To get a token, the CB_API_KEY environnement variable should be a valid API key. Please procure one via the dashboard to continue: https://crunchybridge.com/account/api-keys.\n `
Otherwise, a connected user might assume there is a bug in `cb`, while he is the one having to do something.

abrightwell:
> This seems good to me.
> 
> I have a last minor comment. On `cb token`, I rightfully get the following error:
> 
> ```
> You're using an invalid API key. Please procure a new one via the dashboard to continue.
> ```
> 
> Adding `CB_API_kEY=` works as expected. The wording could however be a little bit improved. Something like: `To get a token, the CB_API_KEY environnement variable should be a valid API key. Please procure one via the dashboard to continue: https://crunchybridge.com/account/api-keys.\n ` Otherwise, a connected user might assume there is a bug in `cb`, while he is the one having to do something.

I've updated the error message to include the URL. The API key can be set both in the credentials file and the ENV. So I didn't mention the ENV itself. We could use some better documentation around this for sure. I'll get some updates into the README before the next release to cover the new login approach, to include the ENV.

abrightwell:
Thanks!

PR #129, opened 2023-06-05
Update README with new login details.

abrightwell:
Updates the README file with the new `login` process. To also include usage of an API Key via the `CB_API_KEY` env.


ngaumont (approved):

abrightwell:
Thanks, I'm going to hold on merging this one until we get closer to the next release. That way there is no potential disconnect between how the current version works.

PR #130, opened 2023-06-05
Add `CREDENTIALS_FILE` constant.

abrightwell:
Refactoring out the credential file path/name to a constant that is based on the configured host (CB_HOST).


abrightwell:
@ngaumont this is a follow up to #128 where we discussed it. Thoughts?

ngaumont (approved):
Thanks!

PR #131, opened 2023-06-08
Set `X-Parent-Request-Id` header.

abrightwell:
Recently the API added support for a `X-Parent-Request-Id` header. This is so it is more apparent that when an a action is taken that requires multiple requests to complete that the multiple requests can be grouped/identified. This is meant to assist in potential debugging and troubleshooting efforts.

So, here we take advantage of this new feature and by setting it when an `APIAction` is initialized so that it will be included on all API requests made by the action.

### DIFF
--- a/src/cb/action.cr
+++ b/src/cb/action.cr
@@ -203,6 +203,10 @@ module CB
     property client : Client
 
     def initialize(@client, @input = STDIN, @output = STDOUT)
+      # Set the parent request id header. This is so that all requests related
+      # to a specific action can be grouped together for logging and debugging
+      # purposes.
+      @client.headers["X-Parent-Request-Id"] = UUID.random.to_s
     end
 
     abstract def run


abrightwell:
@brandur - seem right to you?

brandur:
@abrightwell Hmm, I'm not sure I'd bother setting this unless it's actually doing something useful, like tying multiple requests together.

Just setting it to a random value on every request doesn't really feel that useful IMO — instead, maybe just pull the `Request-Id` off API responses and make that loggable?

abrightwell:
> 



> @abrightwell Hmm, I'm not sure I'd bother setting this unless it's actually doing something useful, like tying multiple requests together.

Hmmm... That's what it should be doing. An APIAction could perform multiple API requests and they are all 'related'. I'm setting it on the client when the action is initialized, so that it will carry through to all API requests that the action makes. It should only be set once per command not per API request.

> Just setting it to a random value on every request doesn't really feel that useful IMO — instead, maybe just pull the `Request-Id` off API responses and make that loggable?

Given the above clarification, does this make sense? Or am I not understanding the purpose/use of the header?

brandur (approved):
OIC, the UUID is only generated once because the client gets reused in case multiple calls are issued.

> Given the above clarification, does this make sense? Or am I not understanding the purpose/use of the header?

Yeah, makes sense. You might want to drop a simple comment on that line to indicate why it's there and why it works since it might not otherwise be super intuitive as to why it's there.

PR #133, opened 2023-07-10
Add `cb config-param` support.

abrightwell:
Add basic support for managing supported cluster configuration parameters.

This command provides the following sub-commands:

```
❯ cb config-param --help
Usage:
    cb config-param <command>

Commands:
    get                              display one or all configuration parameters for a cluster
    list-supported                   display supported configuration parameters
    reset                            reset one or more configuration parameter values to default
    set                              set one or more configuration parameter values

Options:
```


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Install Nix
-        uses: cachix/install-nix-action@v20
+        uses: cachix/install-nix-action@v22
```
abrightwell, line 18: This is unrelated, but necessary to get CI working for `macos`.

spec/cb/login_spec.cr
```diff
@@ -21,6 +21,7 @@ Spectator.describe CB::Login do
     let(client) { mock(Client) }
 
     before_each {
+      ENV["CB_API_KEY"] = nil
```
abrightwell, line 24: Just ensures that it's not set for the purposes of the test. As it's possible that it is set locally and that would create a false positive for the spec. This was missed previously in #128. So, just adding it here since I noticed it during my testing of this feature.

src/models/config_param.cr
```diff
@@ -0,0 +1,17 @@
+module CB::Model
+  jrecord ConfigurationParameter,
+    component : String? = nil,
+    name : String = "",
+    parameter_name : String? = nil,
+    require_restart : Bool = false,
```
abrightwell, line 6: This will become `requires_restart` once the API is updated accordingly. This is the particular item mentioned in the description of this PR.

abrightwell (commented):
@ngaumont thoughts on this one?

src/cb/config_param.cr
```diff
@@ -0,0 +1,181 @@
...
+    # has an effect when the output format is a table.
+    property no_header : Bool = false
+
+    # Result of api calls.
+    property parameters : Array(Model::ConfigurationParameter) = [] of Model::ConfigurationParameter
+
+    # List of configuration parameter argments for the action.
+    property args : Array(String) = [] of String
+
+    def validate
```
ngaumont, line 30: I don't know how much we would like to validate beforehand, but having a minimal validation would help the user getting the right format with a proper error message.
The endgoal would be to warn the user that `postgres_max-connection=100` is not valid because it's not in the format `component:name=value`.

What do you think of adding a dedicated class?
```
class ParameterQuery
  property component : String?
  property name : String?
  property value : String?
end
```

During the validation, the array of string `args` would be transformed to an array `ParameterQuery` with something like this: 
```
args.map{ |arg|
  matched =Regexp.new(/\A(?<component>\w+)?(:(?<param_name>\w+))?(=(?<value>\w+))?\Z/).match(arg)
  raise Error "parameter should be in the" unless matched 
  ParameterQuery.new(matched[:component],matched[:param_name], matched[:value])
```

In each command, we could add more validation to enforce the parameter name or value to b present.

src/cb/config_param.cr
```diff
@@ -0,0 +1,181 @@
...
+    bool_setter allow_restart
+
+    def run
+      super
+
+      parameters = @args.map do |arg|
+        parts = arg.split('=')
+        {"name" => parts[0], "value" => parts[1]}
+      rescue IndexError
+        raise Error.new("Invalid argument: #{arg}). Make sure that it has the following format <name>=<value>.")
```
ngaumont, line 138: See previous comment on validation.
Also the error message could contain more information on the format: `<compontent>:<name>=<value>`

src/cli.cr
```diff
@@ -171,6 +171,96 @@ op = OptionParser.new do |parser|
...
+  # Custom Configuration Parameter Management
+  #
+
+  parser.on("config-param", "Manage configuration parameters") do
+    parser.banner = "cb config-param <command>"
+
+    # TODO (abrightwell): would be really cool to implement this kind of
+    # support.
+    #
+    # parser.on("edit", "interactively edit config values") do
```
ngaumont, line 184: would be awsome. Should be awsome as `\e` does that in psql

src/cli.cr
```diff
@@ -171,6 +171,96 @@ op = OptionParser.new do |parser|
...
+        Get all configuration parameters:
+        $ cb config-params get --cluster <ID>
+      EXAMPLES
+    end
+
+    parser.on("list-supported", "display supported configuration parameters") do
+      parser.banner = "cb config-param list-supported"
+
+      supported = set_action ConfigurationParametListSupported
+      parser.on("--component COMPONENT", "Choose component (default: all)") { |arg| supported.component = arg }
```
ngaumont, line 211: Strange to have `--component` here, but nothing in the other subcommands.
I would expect to use the same args logic as the others.

src/cli.cr
```diff
@@ -171,6 +171,96 @@ op = OptionParser.new do |parser|
...
+      parser.examples = <<-EXAMPLES
+        Revert a single configuration parameter:
+        $ cb config-param revert --cluster <ID> postgres:max_connections
+
+        Revert multiple configuration parameters:
+        $ cb config-param revert --cluster <ID> postgres:max_connections postgres:hot_standby
+
+        Revert all configuration parameters:
+        $ cb config-param revert --cluster <ID> --all
+      EXAMPLES
```
ngaumont, line 260:
```suggestion
        $ cb config-param reset --cluster <ID> postgres:max_connections

        Revert multiple configuration parameters:
        $ cb config-param reset --cluster <ID> postgres:max_connections postgres:hot_standby

        Revert all configuration parameters:
        $ cb config-param reset --cluster <ID> --all
      EXAMPLES
```

src/cli.cr
```diff
@@ -171,6 +171,96 @@ op = OptionParser.new do |parser|
...
+
+      parser.examples = <<-EXAMPLES
+        Revert a single configuration parameter:
+        $ cb config-param revert --cluster <ID> postgres:max_connections
+
+        Revert multiple configuration parameters:
+        $ cb config-param revert --cluster <ID> postgres:max_connections postgres:hot_standby
+
+        Revert all configuration parameters:
+        $ cb config-param revert --cluster <ID> --all
```
ngaumont, line 259: Does `--all` works? I did not expect it to work and find it dangerous.
We might want to allow `cb config-param reset --cluster <ID> postgres`.

ngaumont (changes requested):

src/cb/completion.cr
```diff
@@ -385,6 +388,87 @@ class CB::Completion
...
+    end
+
+    suggest = [] of String
+    suggest << "--cluster\tcluster id" unless has_full_flag? :cluster
+    suggest << "--format\toutput format" unless has_full_flag? :format
+    suggest
+  end
+
+  def config_param_list_supported
+    return [] of String if last_arg?("--component")
```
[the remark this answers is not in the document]
abrightwell: Yeah, I was going to once I knew what the final list would be. Obviously, `postgres` and `pgbouncer` are the two imminent ones, so I could add those without issue, I suppose. My hesitation in providing it statically like this at the moment is that I don't want to have to make a release just to add more components. Though, perhaps this list is stable enough.

Another thought I had was to simply fetch and cache the parameters and filter out the components that way. But then I really didn't like that approach given the multiple/duplicated requests.

I do think, however, there is room for it in future improvements.

I'm open to whatever makes sense here. What do you think?

src/cli.cr
```diff
@@ -171,6 +171,96 @@ op = OptionParser.new do |parser|
...
+      get = set_action ConfigurationParameterGet
+      parser.on("--cluster ID", "Choose cluster") { |arg| get.cluster_id = arg }
+      parser.on("--format FORMAT", "Choose output format (default: table)") { |arg| get.format = arg }
+      parser.on("--no-header", "Do not display table header") { get.no_header = true }
+
+      parser.unknown_args { |args| get.args = args }
+
+      parser.examples = <<-EXAMPLES
+        Get a single configuration parameter:
+        $ cb config-params get --cluster <ID> --name postgres:max_connections
```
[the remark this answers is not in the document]
abrightwell: Ah! Good catch. Thanks!

src/cb/config_param.cr
```diff
@@ -0,0 +1,181 @@
...
+
+  class ConfigurationParametListSupported < ConfigurationParameterAction
+    property component : Component = Component::All
+
+    def component=(str : String)
+      @component = Component.parse(str)
+    end
+
+    def run
+      parameters = client.list_supported_configuration_parameters
```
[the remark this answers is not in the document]
abrightwell: Honestly, I'm not sure. I think given the nature of it that perhaps a `query_param` would be better? But I wouldn't be opposed to either approach. I figured it was just as easy to fetch and filter in the client for the time being. But if the API would like to support this at some point, it might be nice-to-have. 🤷‍♂️ Let me think on it.

src/cb/config_param.cr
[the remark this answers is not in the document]
abrightwell: In short, yes, that's on my list of future improvements.

Longer. Being able to suggest would be really nice. Though, I want to get some kind of caching in place for that as I don't want to have to call out every time to the `supported` endpoint to get the list for each completion.

src/cli.cr
abrightwell: 🤦‍♂️ Thanks!

src/cli.cr
abrightwell: Can you describe what you mean? I'm not sure I follow.

src/cli.cr
abrightwell: No. It does not. I intentionally did not implement for the obviously dangerous consequences of it. Again, another `example` I need to update.

src/cli.cr
ngaumont: I expected to be able to write:
```
# to get all the param
cb config-param list-supported
# to get only postgres parameters
cb config-param list-supported postgres
```

src/cb/completion.cr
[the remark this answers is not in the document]
ngaumont: For a v1, I won't bother to try to suggest anything. I don't expect `cb config-param list-supported` to be heavily used.
`list-supported` could do even less validation and don't try to validate component name at all.
I hope that the error from the API could be good enough, at least for v1.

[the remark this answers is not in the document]
abrightwell: Yeah, I'm going to update it to not validate the component name. Since the filtering is happening client side, I'm not too concerned about an invalid value at this point as it would just return an empty list of parameters. I feel like that is safe enough for now. As well, it keeps us from having to worry too much about updating just to support a new component, when really no update should be necessary for such a thing.

src/cb/config_param.cr
abrightwell: So, I'll need to think on this one... we at least have some basic validation going on that indicates that the argument given is invalid:

```
❯ ./bin/cb config-param set --cluster dy4kh7kr3bavriw5pmihta6cdy postgres_log_lock_waits=true
error: Configuration parameters not supported through the API: "postgres_log_lock_waits"
```

But I'd agree more information on why it is invalid would be nice-to-have. I'll chew on it.

abrightwell: I could probably add a message that to the end of this one that says something like:

`ensure the input is in the following format component:parameter=value` or something similar.

abrightwell:
@ngaumont mind taking another look at this one? 🙂

src/cb/config_param.cr
```diff
@@ -0,0 +1,160 @@
...
+  class ConfigurationParameterGet < ConfigurationParameterAction
+    def run
+      super
+
+      if @args.empty?
+        parameters = client.list_configuration_parameters(cluster_id[:cluster])
+      else
+        name = @args.try &.first
+        parameters = [client.get_configuration_parameter(cluster_id[:cluster], name.to_s)]
+      end
```
ngaumont, line 73: `name` is already a String and I suppose `.to_s` is not necessary.

```suggestion

      if (name = @args.try &.first)
        parameters = [client.get_configuration_parameter(cluster_id[:cluster], name)]
      else
        parameters = client.list_configuration_parameters(cluster_id[:cluster])
      end
```

src/cb/config_param.cr
```diff
@@ -0,0 +1,160 @@
...
+    bool_setter allow_restart
+
+    def run
+      super
+
+      updated_parameters = @args.map do |arg|
+        parts = arg.split('=')
+        {"name" => parts[0], "value" => parts[1]}
+      rescue IndexError
+        raise Error.new("Invalid argument: #{arg}). Make sure that it has the following format <component>:<name>=<value>.")
```
ngaumont, line 127: This is fine like this, but could be better in the validation step (for the next PR).

ngaumont (approved):
Few minor comments (typo or small improvement), but can be merged once they are addressed.

src/cb/config_param.cr
```diff
@@ -0,0 +1,160 @@
...
+
+    protected def output_json(parameters : Array(Model::ConfigurationParameter))
+      output << {
+        "parameters": parameters,
+      }.to_pretty_json << '\n'
+    end
+  end
+
+  # Action for getting configuration parameters.
+  class ConfigurationParameterGet < ConfigurationParameterAction
```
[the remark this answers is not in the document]
abrightwell: Yeah, my original thought was to just ignore anything except the first... but agreed, having a validation would be more appropriate.

src/cb/config_param.cr
abrightwell: Good catch! 🙏  Previous implementation had `name : String?` as it was a separate property set by `--name`. So the `.to_s` was necessary here. But no longer since we're now just pulling it from `@args`.

src/cb/config_param.cr
abrightwell: Yeah, I had considered that as well. The initial reason that I didn't was that I just didn't find it beneficial at that step as I didn't want to 1) reiterate the list in `#run` again to parse the values and 2) I didn't want to introduce a new `property` to the class that could be shared between `#validate` and `#run`. I could certainly be convinced of `#2`. I'll chew on it and propose something different later.

PR #136, opened 2023-08-10
Fix input validation on `cb create [--fork | --replica]`

abrightwell:
It was found that input validation was failing when creating a fork or
replica using the `--network` option. Specifically, the following error
was being returned:

```
error: Missing required arguments: platform, plan, region, storage
```

This is because when specifying a network none of these items are
required. Most importantly the platform and region are not allowed. This
was a minor oversight when allowing cross-region replicas was first
implemented.  Here we simply fix that issue and if `--network` is given
then we do not require the platform or region as well.


abrightwell (commented):
@ngaumont mind taking a look at this one when you have a chance? Thanks!

PR #141, opened 2023-11-28
specs: add look_like matcher

will:
This makes it so the whitespace can be trimmed from the spec files and not break the tests, and also remove the exception from ameba


abrightwell (approved):
👍

PR #143, opened 2023-11-29
Change from size to storage on hints

Winslett:
With the latest release, the long argument name is "storage".


abrightwell (approved):
Thanks. 👍

PR #153, opened 2024-01-26
Add `cb psql` query menu.

abrightwell:
This is an initial phase of supporting a builtin menu of useful queries when using `cb psql`. Currently, it only supports predefined queries. To access this menu, simply enter `:menu` into psql and it will present the available query options.

Example:

```
Example Team/example-cluster/postgres=> :menu
Cache
  1 – Cache and index hit rates
Size Information
  2 – Database sizes
  3 – Table sizes
Query Performance
  4 – Queries consuming the most system time
  5 – Queries running over 1 minute
  6 – Slowest average queries
Connection Management
  7 – Connection count by state
  8 – Connection count by user and application
Indexes
  9 – Duplicate indexes
  10 – List of indexes
  11 – Unused indexes
Locks
  12 – Blocking queries
Extensions
  13 – Available extensions
  14 – Installed extensions

Type choice and press <Enter> (q to quit):

```

### DIFF
--- a/src/cb/query_menu/menu.psql.ecr
+++ b/src/cb/query_menu/menu.psql.ecr
@@ -1,0 +1,32 @@
+<%- option : Int32 = 1 %>
+<%- @queries.each do |category, queries| -%>
+\echo <%= "#{category.colorize.bold}" %>
+<%- queries.each do |query| -%>
+    \echo '  <%= "#{option} \u2013 #{query.label}" %>'
+    <%- option += 1 -%>
+<%- end -%>
+<%- end -%>
+\echo
+\prompt 'Type choice and press <%= "<Enter>".colorize.bold %> (<%= "q".colorize.bold %> to quit): ' choice
+\echo
+
+<%- option = 1 %>
+SELECT CASE
+<%- @queries.each_value do |queries| -%>
+<%- queries.each do |query| %>
+WHEN :'choice'::text = '<%= option %>' THEN
+'\i `echo <%= query.path %>`'
+'\echo'
+'\i <%= @path %>'
+<%- option += 1 -%>
+<%- end -%>
+<%- end -%>
+WHEN :'choice'::text = 'q'
+THEN '\echo Quitting!'
+ELSE
+'\echo <%= "Error:".colorize.red.bold %> Unknown option! Try again.'
+'\echo'
+'\i <%= @path %>'
+END AS action \gset
+
+:action
--- a/src/cb/query_menu/query.cr
+++ b/src/cb/query_menu/query.cr
@@ -1,0 +1,89 @@
+require "base64"
+
+module CB::QueryMenu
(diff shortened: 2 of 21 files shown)


src/cb/query_menu/menu.psql.ecr
```diff
@@ -0,0 +1,32 @@
...
+\echo
+
+<%- option = 1 %>
+SELECT CASE
+<%- @queries.each_value do |queries| -%>
+<%- queries.each do |query| %>
+WHEN :'choice'::text = '<%= option %>' THEN
+'\i `echo <%= query.path %>`'
+'\echo'
+'\i <%= @path %>'
```
ngaumont, line 20: I'm not sure, we want to loop after a successful query.
I find it that is burying the useful information.

Ideally, I would prefer to have `:menu` display the options only once and if we do loop add the entry `?`,`h` or `help` to list possible entry.

Also, I would love if after entry I would be able to `\e` to see the actual query.

src/cb/query_menu/query_menu.cr
```diff
@@ -0,0 +1,32 @@
...
+      # Write the queries to the filesystem.
+      @queries.each_value { |queries| queries.each(&.write) }
+
+      # Render the menu file.
+      @path = File.join(temp_dir, "menu.psql")
+      query_menu = File.open(@path, mode: "w") do |menu|
+        menu << ECR.render __DIR__ + "/menu.psql.ecr"
+      end
+
+      "\\set menu '\\\\i #{query_menu.path} '"
```
ngaumont, line 29: Maybe advertise that `:menu` even exists:

```suggestion
      "\\set menu '\\\\i #{query_menu.path} '"
      "\\echo type #{:menu.colorize.bold} to execute commonly useful queries "
```

src/cb/query_menu/query.cr
```diff
@@ -0,0 +1,89 @@
+require "base64"
+
+module CB::QueryMenu
+  # Metadata annoation for a query.
+  #
+  # A `Query` must have a `label` and `category` field defined using this
+  # annotation.
+  annotation Metadata; end
```
ngaumont, line 8: I'm completely new to annotation. So, this is more a request to learn than a comment.
Could the Metadata be removed in favor of a standard `Query` class stating that each instance should have a `label`, `category` and `raw_sql_path`?

`Query.all` would hold the list, and would then be something like :

```crystal
[
  Query.new(label: "Table sizes", category: CATEGORY_SIZE_INFORMATION), raw_sql_path("#{__DIR__}/sql/size_information_table_size.sql"),
  …
]
```

src/cb/query_menu/query_menu.cr
abrightwell: I've added this to the `psqlrc` rendering that occurs in `psql.cr`. 👍

PR #154, opened 2024-01-30
Deduced frontend host from Host

ngaumont:


src/cb.cr
```diff
@@ -12,6 +12,8 @@ module CB
 
   HOST = ENV["CB_HOST"]? || "api.crunchybridge.com"
 
+  FRONTEND_HOST = ENV["CB_FRONTEND_HOST"]? || HOST.gsub(/\Aapi\./, "")
```
brandur, line 15: Minor, but I wonder if we should be using `DASHBOARD` instead of `FRONTEND`? The former is how it's referenced in documentation, etc. I usually think of it as the "Frontend" team and the "Dashboard" product.

abrightwell: Yeah, I'm in agreement. I'd prefer `DASHBOARD` as well.

src/cb/login.cr
```diff
@@ -60,7 +60,7 @@ module CB
       # Request a session intent.
       si_params = Client::SessionIntentCreateParams.new(agent_name: "cb #{CB::VERSION}")
       session_intent = @client.create_session_intent si_params
-      login_url = "https://www.crunchybridge.com/account/verify-cli/#{session_intent.id}?code=#{session_intent.code}"
+      login_url = "https://#{FRONTEND_HOST}/account/verify-cli/#{session_intent.id}?code=#{session_intent.code}"
```
jaidetree, line 63: Is it possible the DASHBOARD_HOST or FRONTEND_HOST var could also contain the protocol? I am not able to test this locally

jaidetree: I can try setting up a ngrok proxy which might work, but not ideal

abrightwell: 🤔 I'd prefer not to mix the host portion of the URL with other parts. Let me think about how we might be able to solve for that particular case though as I understand the issue.

ngaumont: I updated to `DASHBORD_HOST`, but let @abrightwell judge on how we might want to tackle the localhost issue.

src/cb.cr
```diff
@@ -12,6 +12,8 @@ module CB
 
   HOST = ENV["CB_HOST"]? || "api.crunchybridge.com"
 
+  DASHBOARD = ENV["CB_DASHBOARD_HOST"]? || HOST.gsub(/\Aapi\./, "")
```
abrightwell, line 15:
```suggestion
  DASHBOARD_HOST = ENV["CB_DASHBOARD_HOST"]? || HOST.gsub(/\Aapi\./, "")
```

abrightwell (approved):
Just one minor change/fix. Otherwise LGTM. 👍

PR #156, opened 2024-03-28
Update to a newer version of our crystal flake

will:
The main nixpkgs repo is doing a better job of staying up-to-date with crystal releases, so I've removed most of what we were doing to build our own crystal compiler, and drop down to just some quality-of-life wrapping with common libraries.

Also now updating cb to use newer versions can happen by only updating the nixpkgs reference here, since the crunchydata/nixpkgs flake can be told to follow that input.

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -3,6 +3,10 @@
 # The point is for the user to remove these configuration records
 # one by one as the reported problems are removed from the code base.
 
+# Seems broken in current (v1.6.1) build of ameba, so disabling
+Lint/UselessAssign:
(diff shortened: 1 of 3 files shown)


.ameba.yml
```diff
@@ -3,6 +3,10 @@
 # The point is for the user to remove these configuration records
 # one by one as the reported problems are removed from the code base.
 
+# Seems broken in current (v1.6.1) build of ameba, so disabling
```
abrightwell, line 6: 👍

abrightwell (approved):

PR #162, opened 2024-05-08
Fix `cb uri` hostname for replicas using `user` role.

abrightwell:
It was reported that using `cb uri` to retrieve credentials for a `user` role on a replica was returning the hostname of the `primary` instead of the target replica.

This was because it's possible that the` user` role does not exist yet and therefore would need to first be created on the primary. So, to ensure that it is, we always attempt to create it on the primary and then retrieve it for the replica. Previously, we were only creating it and that was resulting in the returned credentials to be associated with the primary. So, we've now eliminated that branch in the code so that it will always retrieve this information from the correct target.


cyberdelia (approved):

abrightwell:
Thanks!

PR #163, opened 2024-05-08
Fix config param valus being truncated.

abrightwell:
It was reported that the `cb config-param set` command was truncating values that contain a `=` character. This was due to how the input line was being parsed and tokenized. Here, we've simply limited the parsing to the first instance of this character such that it will only ever split the parameter name and value from the input.


brandur (approved):

abrightwell:
Thanks!

PR #169, opened 2024-05-29
Add support for managing VPC peering.

abrightwell:
Add support for managing VPC peering via the `cb network` command.

These changes provide the following subcommands:

* `cb network create-peering` - create a new VPC peering in the target network.
* `cb network delete-peering` - delete an existing VPC peering.
* `cb network get-peering` - show details of an existing VPC peering.
* `cb network list-peerings` - list all existing VPC peerings for the target network.

### DIFF
--- a/src/cb/peering.cr
+++ b/src/cb/peering.cr
@@ -1,0 +1,178 @@
+require "./action"
+
+module CB
+  abstract class PeeringAction < APIAction
+    format_setter format
+    eid_setter network_id
+    property? no_header : Bool = false
+
+    abstract def run
+
+    def validate
+      check_required_args do |missing|
+        missing << "network" unless @network_id
+      end
+    end
+
+    def display(peerings : Array(Model::Peering))
+      case @format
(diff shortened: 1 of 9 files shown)


src/cb/peering.cr
```diff
@@ -0,0 +1,166 @@
...
+  class PeeringCreate < PeeringAction
+    property platform : String?
+    property aws_account_id : String?
+    property aws_vpc_id : String?
+    property gcp_project_id : String?
+    property gcp_vpc_name : String?
+
+    def validate
+      super
+      check_required_args do |missing|
```
ngaumont, line 97: should we try to proactively prevent user to use aws_account_id if platform is not aws (same with gcp, etc)?

ngaumont (approved):
Very minor, remark, but this seems good to me

src/cb/peering.cr
abrightwell: Good catch. Yeah, we definitely should do that kind of check here. 👍

PR #170, opened 2024-07-25
Fix expired session login redirect.

abrightwell:
When we switched to session based authentication from token based, we
introduced an undesired UX issue when a session expires. In summary,
when a session expires, the desired functionality is that the user will
be presented with a message indicating the expiration and prompted to
login in again.

While the first of these was occuring in the form of a formatted HTTP
error output the latter was not. Also, the form of the error was less
than desirable too.

Therefore, what we've done here is updated the error handling around
session expiration/unauthorized errors to prompt the user to login
again.  Also, we've taken the opporunity to clean up the error output of
other API errors as well such that it's presents the error. Use of
`HTTP_DEBUG` is still supported if the full HTTP error details is
necessary for debugging purposes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `cb info` now displays `tailscale` status.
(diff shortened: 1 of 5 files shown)


ngaumont (approved):

abrightwell:
Thanks.

PR #174, opened 2024-10-17
Actually use user's ~/.psqlrc file

rafbm:
Hey! Happy Bridge customer here. 👋😄

Since as long as I remember, my `~/.psqlrc` customizations have never been available in `cb psql` sessions. Despite [this commit](https://github.com/CrunchyData/bridge-cli/commit/0a2211b78b203a72d3608a06aee519ddd6570c05), for instance, I think this probably never worked? Related changes in Crystal behavior seem to go [a while back](https://github.com/crystal-lang/crystal/commit/b98c0ffa82e78b4c3b586fdcb985c145417d3547).

Here’s the fix, confirmed to work on my machine. (It’s awesome being able to test by just swapping `cb` for `src/cli.cr`!)

Given specs were green and still are, I assume that part just isn’t tested so didn’t study the test suite to find where to add it.

### DIFF
--- a/src/cb.cr
+++ b/src/cb.cr
@@ -14,6 +14,8 @@ module CB
 
   DASHBOARD_HOST = ENV["CB_DASHBOARD_HOST"]? || HOST.gsub(/\Aapi\./, "")
 
+  PSQLRC = File.expand_path(ENV["PSQLRC"]? || "~/.psqlrc", home: true)
+
   # Release constants.
   BUILD_RELEASE  = {{ flag?(:release) }}
   SHARDS_VERSION = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
--- a/src/cb/psql.cr
+++ b/src/cb/psql.cr
@@ -95,7 +95,7 @@ module CB
       end
 
       psqlrc = File.tempfile(c.id, "psqlrc")
-      File.copy("~/.psqlrc", psqlrc.path) if File.exists?("~/.psqlrc")
+      File.copy(CB::PSQLRC, psqlrc.path) if File.exists?(CB::PSQLRC)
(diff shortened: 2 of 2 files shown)


abrightwell:
Hey @rafbm, thanks for the PR. I'll work on including it into the next release.

While I'm sure it is less common, it seems that we **might** want to also consider what might happen if the user were to make use of the [PSQLRC](https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-ENVIRONMENT-PSQLRC) environment variable. More specifically, if the location is outside of the user's home directory. I'll poke at it and let you know if I think we should include here.

I'm not terribly concerned with handling the [version specific](https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-FILES-PSQLRC) convention/naming at this time. But perhaps that's something to think about as well, **just** in case. I'll give that one more thought for later.

Also, this made me curious as to handling the `-X, --no-psqlrc` option if it were provided given: `cb psql <ID> -- -X`. Some initial testing seems to suggest that it's not a concern. And it's further supported given that `psql` will ignore loading/processing `.psqlrc` when this option is given which is confirmed [here](https://github.com/postgres/postgres/blob/98c7c7152d2d4670a349c0fe48f05cf95c0b1185/src/bin/psql/startup.c#L331-L332) and [here](https://github.com/postgres/postgres/blob/98c7c7152d2d4670a349c0fe48f05cf95c0b1185/src/bin/psql/startup.c#L347-L348).

rafbm:
Great points! Feel free to amend this PR in any way. This was my first interaction with Crystal so you’ll sure be faster than me at any further improvements. 😄

abrightwell (approved):

PR #179, opened 2025-04-01
Update cluster `state` enum with new values.

abrightwell:
Updates the cluster `state` enum with recently exposed state values.
This is currently an exhaustive list of possible cluster states.


brandur (approved):

abrightwell:
Thanks.

PR #180, opened 2025-04-03
Allow `cb open` redirect to be modified by `CB_REDIRECT_URL`

brandur:
Here, let `cb open` redirect to a non-standard Dashboard URL (or other
internal component) by respecting the `CB_REDIRECT_URL` env var. If the
var's empty, it continues to have the same behavior as before.


brandur:
@abrightwell Finally got the tests fixed up here. Mind taking a look?

abrightwell (approved):

PR #191, opened 2026-05-04
Add saved query management functions.

sfc-gh-jsoyland:


sfc-gh-jsoyland:
Tried a 3-phased approach to this PR with checkpoints that I could review.  I also found that it made the process repeatable on a fresh branch.

1. First asked Claude to research how to best apply a new function by reviewing existing code and API documentation.  Then to persist findings to [01-research.md](https://github.com/user-attachments/files/27370355/01-research.md) for review.

2. Asked Claude to plan what changes would be needed based on findings from the research document:  [02-planning.md](https://github.com/user-attachments/files/27370356/02-planning.md)

3. Asked Claude to produce a testing plan for review post-implementation.  That way I could review the doc and make sure that it wasn't trying to cheat in it's success criteria:  [03-testing.md](https://github.com/user-attachments/files/27370357/03-testing.md)

Once artifacts were generated, started a new session and asked Claude to implement changes.  Everything generated/passed on the first pass, save for the linter error.

src/cb/saved_query.cr
```diff
@@ -0,0 +1,121 @@
...
+        missing << "query" unless query_id
+      end
+    end
+
+    def run
+      validate
+      query = client.get_saved_query query_id
+
+      filename = @file || "#{query.name.gsub(/[^a-zA-Z0-9_\-]/, "_")}.sql"
+      File.write(filename, query.sql)
```
sfc-gh-abrightwell: Maybe have a `rescue` block in case this fails so that the user gets an error message over a stack trace?

src/cb/saved_query.cr
```diff
@@ -0,0 +1,121 @@
...
+      check_required_args do |missing|
+        missing << "cluster" unless cluster_id
+        missing << "file" unless file
+        missing << "name" unless name
+      end
+    end
+
+    def run
+      validate
+      sql = File.read(@file.to_s)
```
sfc-gh-abrightwell: Same here maybe, if the file doesn't exist?

Spitballing, but could maybe do something like:

```
unless File.exists?(@file.to_s) <some error>?
```

Maybe crystal does something reasonable though if not? Maybe worth validating? 🤷‍♂️

spec/cb/saved_query_spec.cr
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  describe "#run" do
+    before_each do
+      action.cluster_id = "pkdpq6yynjgjbps4otxd7il2u4"
+      action.query_id = "sqpvoqooxzdrriu6w3bhqo55c4"
+    end
+
+    it "exports to specified file" do
+      action.file = "/tmp/test_export.sql"
```
sfc-gh-abrightwell: Might be worth considering using [Dir.tempdir](https://crystal-lang.org/api/1.20.2/Dir.html#tempdir%3AString-class-method)? As it abstracts the system tempdir which I think is `/var/folders` on macos and `/tmp` on Linux... 🤔

src/client/saved_query.cr
```diff
@@ -0,0 +1,43 @@
...
+
+      saved_queries
+    end
+
+    def get_saved_query(saved_query_id)
+      resp = get "saved-queries/#{saved_query_id}"
+      CB::Model::SavedQuery.from_json resp.body
+    end
+
+    def create_saved_query(params)
```
sfc-gh-abrightwell: It might be worth definite a param type for the create method. As I *think* that leaving it unbounded chokes spectator when generating the mock client. Or at least that's one thing I feel like eased the CI failures a little that I was debugging in #192.

```
    struct SavedQueryCreateParams
      include JSON::Serializable
      property cluster_id : String?
      property name : String?
      property sql : String?
      property? skip_enqueue : Bool = true

      def initialize(@cluster_id, @name, @sql, @skip_enqueue = true)
      end
    end

    def create_saved_query(params : SavedQueryCreateParams)
      resp = post "saved-queries", params
      CB::Model::SavedQuery.from_json resp.body
    end
```

sfc-gh-abrightwell (commented):
LGTM. Just a few comments.

src/cb/saved_query.cr
sfc-gh-jsoyland: 👍 Added code to capture error:
```
      begin
        File.write(filename, query.sql)
      rescue e : IO::Error
        raise Error.new "Failed to write file '#{filename}': #{e.message}"
      end
```

src/cb/saved_query.cr
sfc-gh-jsoyland: Add code to error if file doesn't exist:
```
      begin
        sql = File.read(@file.to_s)
      rescue e : IO::Error
        raise Error.new "Failed to read file '#{@file}': #{e.message}"
      end
```

spec/cb/saved_query_spec.cr
sfc-gh-jsoyland: Good idea; changed to use `Dir.tempdir` here.

sfc-gh-jsoyland:
Ran through execution paths that were updated (temp directory for export, error handling bad file, and new type being used as a parameter):
```shell
$ ./cb saved-query list --cluster mlfoijrgqbbjxmnngd6xrgdn64
  ID                           Name                  Query
  agpglfhr6z6m3r6nch32zvy23q   Untitled Query 99f5   select 1;
$  ./cb saved-query export --cluster mlfoijrgqbbjxmnngd6xrgdn64 --query agpglfhr6z6m3r6nch32zvy23q
exported Untitled Query 99f5 to Untitled_Query_99f5.sql
$ cat Untitled_Query_99f5.sql
select 1;⏎
$ mv Untitled_Query_99f5.sql  select.sql
$ ./cb saved-query import --cluster mlfoijrgqbbjxmnngd6xrgdn64 --file asdf.sql --name "jesse_test_query"
error: Failed to read file 'asdf.sql': Error opening file with mode 'r': 'asdf.sql': No such file or directory
$ ./cb saved-query import --cluster mlfoijrgqbbjxmnngd6xrgdn64 --file select.sql --name "jesse_test_query"
created saved query agpglfvq2fufsgl7c2ez3l3zi4
$ ./cb saved-query list --cluster mlfoijrgqbbjxmnngd6xrgdn64
  ID                           Name                  Query
  agpglfvq2fufsgl7c2ez3l3zi4   jesse_test_query      select 1;
  agpglfhr6z6m3r6nch32zvy23q   Untitled Query 99f5   select 1;
```

sfc-gh-abrightwell (approved):

